PR #3, opened 2023-07-18
feat: Systems data source template

ggracechoi:
Created new plugin skeleton for the systems data source repository using _npm run generate_ ([link to user story](https://dev.azure.com/ni/DevCentral/_workitems/edit/2453814))


src/datasources/system/README.md
```diff
@@ -0,0 +1,6 @@
+# Systemlink System data source
+
+<!--
+    TODO: Write a short description of your plugin and document any extra
+    configuration that is required for development.
+-->
```
mure, line 6: Let's fill out the description here so we don't forget to do it later on. Doesn't need to be too involved - you can just write an overview of the plugin's features.

mure (changes requested):

src/datasources/data-frame/README.md
```diff
@@ -1,6 +1,11 @@
...
--->
+This plugin will have three main workflows:
+
+- Populating a
+  [query variable](https://grafana.com/docs/grafana/latest/dashboards/variables/add-template-variables/#add-a-query-variable)
+  with a list of user-accessible systems
+
+- Visualizing system connection status counts on a dashboard
+
+- Visualizing system metadata on a dashboard
```
mure, line 11: This is the wrong plugin's README.

Also, let's use the present tense here. Maybe something like:

```suggestion
This is a plugin for the Systems Management service. It allows you to:

- Populate a
  [query variable](https://grafana.com/docs/grafana/latest/dashboards/variables/add-template-variables/#add-a-query-variable)
  with a list of systems connected to the server
- Visualize system connection status counts on a dashboard
- Visualize system metadata on a dashboard
```

mure (approved):

PR #4, opened 2023-07-20
feat: implement testDatasource method for systems data source

ggracechoi:
# Justification: 
A health check needs to be implemented to see if a response can be retrieved using the provided URL. A valid response would indicate that the configuration is correct.

# Implementation: 
The testDataSource function was implemented using the _/get-systems-summary_ route. A response containing the status counts indicates a valid response for this check. Additionally, a confirmation message displays upon success.

# Testing: 
The web dev tools were used to test this change. The response was checked to see if the expected response was returned by the query.


mure (approved):

mure:
Don't forget to prefix your PR title with `feat: ` so that it triggers a new release. Also, my preference is to have PR titles/commit messages be more concrete instead of being named after the user story. So something like:
> feat: implement `testDatasource` method for systems data source

PR #6, opened 2023-07-24
feat: display system status counts

ggracechoi:
# Pull Request

## 🤨 Rationale

The _Summary_ query type needs to display the status counts of the systems (_Connected_ and _Disconnected_) on a dashboard. This PR satisfies [this user story](https://dev.azure.com/ni/DevCentral/_workitems/edit/2453817).

## 👩‍💻 Implementation

The _Connected_ and _Disconnected_ counts are pulled from the _/get-systems-summary_ route. An _InflineFieldRow_ form and a _RadioButtonGroup_ form were added to create the visual components in the query editor. A new type called _SystemSummary_ was created, and the _onQueryTypeChange_ method and the _query_ method were implemented. 

## 🧪 Testing

The web dev tools were used to test this change. We checked to see if a call was being made when the _Summary_ tab was clicked and that the response matched what was expected. 

## ✅ Checklist

- [ x ] This PR has a title that follows the [commit message format](../README.md#commit-message-format).


cameronwaterman:
@mure - The checklist link in the PR description is broken FYI

src/datasources/system/SystemDataSource.ts
```diff
@@ -24,15 +24,26 @@ export class SystemDataSource extends DataSourceApi<SystemQuery> {
...
+            { name: 'Connected Count', values: [summaryResponse.connectedCount], type: FieldType.number },
+            { name: 'Disconnected Count', values: [summaryResponse.disconnectedCount], type: FieldType.number },
+          ],
+        });
+      } else {
+        return new MutableDataFrame({
+          refId: target.refId,
+          fields: [
+            { name: 'Time', values: [from, to], type: FieldType.time },
+            { name: 'Value', values: [1, 2], type: FieldType.number },
```
cameronwaterman, line 42: What are the hardcoded 1 and 2 representative of? Can we pull these out as constants with a descriptive name?

src/datasources/system/SystemDataSource.ts
```diff
@@ -24,15 +24,26 @@ export class SystemDataSource extends DataSourceApi<SystemQuery> {
...
-        ],
-      });
-    });
+    const data = await Promise.all(options.targets.map(async (target) => {
+      if (target.queryClass === QueryType.Summary) {
+        var summaryResponse = await getBackendSrv().get<SystemSummary>(this.baseUrl + '/get-systems-summary');
+        return new MutableDataFrame({
+          refId: target.refId,
+          fields: [
+            { name: 'Connected Count', values: [summaryResponse.connectedCount], type: FieldType.number },
```
cameronwaterman, line 33: @mure Do we have a precedent for using title case or sentence case here?

src/datasources/system/components/SystemQueryEditor.tsx
```diff
@@ -1,32 +1,30 @@
...
   return (
-    <div className="gf-form">
-      <InlineField label="Constant">
-        <Input onChange={onConstantChange} value={constant} width={8} type="number" step="0.1" />
-      </InlineField>
-      <InlineField label="Query Text" labelWidth={16} tooltip="Not used yet">
-        <Input onChange={onQueryTextChange} value={queryText || ''} />
-      </InlineField>
+    <div>
+      <InlineFieldRow >
```
cameronwaterman, line 23:
```suggestion
      <InlineFieldRow>
```

src/datasources/system/types.ts
```diff
@@ -1,6 +1,15 @@
...
+
 export interface SystemQuery extends DataQuery {
-  queryText?: string;
-  constant: number;
+  queryClass: QueryType
+}
+
+export interface SystemSummary {
+  connectedCount: number,
+	disconnectedCount: number
```
cameronwaterman, line 14:
```suggestion
  disconnectedCount: number
```

src/datasources/system/SystemDataSource.ts
ggracechoi: Oh everything in the else statement is just a placeholder from the template. I can change it to return null for now if that's better?

src/datasources/system/types.ts
```diff
@@ -1,6 +1,15 @@
 import { DataQuery } from '@grafana/schema'
 
+export enum QueryType {
+  Metadata = "Metadata",
```
cameronwaterman, line 4: Do we actually need to reassign the default enum values here?

cameronwaterman (approved):

src/datasources/system/types.ts
ggracechoi: I think Carson just said it was cleaner that way

src/datasources/system/SystemDataSource.ts
cameronwaterman: If the placeholder values truly serve no purpose, then yes I'd suggest we don't set a value instead.

src/datasources/system/types.ts
mure: If we don't explicitly set a value here, then the query type would just get saved as integers.

src/datasources/system/components/SystemQueryEditor.tsx
```diff
@@ -1,32 +1,30 @@
...
 import { SystemDataSource } from '../SystemDataSource';
-import { SystemQuery } from '../types';
+import { QueryType, SystemQuery } from '../types';
 
 type Props = QueryEditorProps<SystemDataSource, SystemQuery>;
 
+const QUERY_TYPES = [
+  {label: "Metadata", value: QueryType.Metadata},
+  {label: "Summary", value: QueryType.Summary}
+]
```
mure, line 12: It would nice if we had a generic utility method for generating this structure from an enum. Can you create that function in a new file `src/core/utils.ts`? The function signature will be:

```ts
export function enumToOptions(stringEnum: { [name: string]: string }): SelectableValue<string>
```

You should be able to get an array of name/value pairs from an enum using `Object.entries`, and then you can map that into to objects of `{ label, value }`.

src/datasources/system/components/SystemQueryEditor.tsx
```diff
@@ -1,32 +1,30 @@
...
-      <InlineField label="Constant">
-        <Input onChange={onConstantChange} value={constant} width={8} type="number" step="0.1" />
-      </InlineField>
-      <InlineField label="Query Text" labelWidth={16} tooltip="Not used yet">
-        <Input onChange={onQueryTextChange} value={queryText || ''} />
-      </InlineField>
+    <div>
+      <InlineFieldRow>
+        <InlineField label="Query Type">
+          <RadioButtonGroup options={QUERY_TYPES} onChange={onQueryTypeChange} value={query.queryClass} />
```
mure, line 25: Once we have that new utility function, this can be

```suggestion
          <RadioButtonGroup options={enumToOptions(QueryType)} onChange={onQueryTypeChange} value={query.queryClass} />
```

mure (changes requested):
Looks like you've got a lint error in the build

src/datasources/system/types.ts
```diff
@@ -1,6 +1,15 @@
...
+  Summary = "Summary"
+}
+
 export interface SystemQuery extends DataQuery {
-  queryText?: string;
-  constant: number;
+  queryClass: QueryType
+}
+
```
mure, line 11: It's good practice to have a default query populated as soon as the user opens the editor instead of having nothing selected:
image

You can solve this by implementing the `getDefaultQuery` method. See the data frame data source for an example:

https://github.com/ni/systemlink-grafana-plugins/blob/2b21199edd6ce17caa86b65fadd213aafbda75f3/src/datasources/data-frame/DataFrameDataSource.ts#L77-L79

(I actually took a look at the grafana source, and the `_.clone` is overkill, so you can omit it). I think "Summary" makes sense as the default type.

src/datasources/system/components/SystemQueryEditor.tsx
```diff
@@ -1,32 +1,30 @@
...
-    <div className="gf-form">
-      <InlineField label="Constant">
-        <Input onChange={onConstantChange} value={constant} width={8} type="number" step="0.1" />
-      </InlineField>
-      <InlineField label="Query Text" labelWidth={16} tooltip="Not used yet">
-        <Input onChange={onQueryTextChange} value={queryText || ''} />
-      </InlineField>
+    <div>
+      <InlineFieldRow>
+        <InlineField label="Query Type">
```
mure, line 24: The label here should be sentence case to match our other plugins:

```suggestion
        <InlineField label="Query type">
```

mure (changes requested):

mure (approved):

PR #7, opened 2023-07-26
feat: display system metadata

ggracechoi:
# Pull Request

## 🤨 Rationale

The _Metadata_ query type needs to display the metadata of all accessible systems on a dashboard. The following fields need to be included: 

- id
- alias
- connection status
- locked status
- system start time
- model
- vendor
- operating system
- ip address
- workspace

This PR satisfies [this user story](https://dev.azure.com/ni/DevCentral/_workitems/edit/2453819).

## 👩‍💻 Implementation

A new _SystemInfo_ interface was created in _types.ts_ to encapsulate all of the different fields. The metadata was then pulled from the _/query-systems_ route in _SystemDataSource.ts_. Due to the large amount of data in the response, several projections were made to extract the specific data that was needed. 

## 🧪 Testing

The web dev tools were used to test this change. We checked to see if a call was being made when the _Metadata_ tab was clicked and that the response matched what was expected.

## ✅ Checklist

- [ X ] This PR has a title that follows the [commit message format](https://github.com/ni/systemlink-grafana-plugins#commit-message-format).


src/core/utils.ts
```diff
@@ -0,0 +1,11 @@
+import { SelectableValue } from "@grafana/data";
+
+export function enumToOptions<T>(stringEnum: { [name: string]: T }): Array<SelectableValue<T>> {
+    const RESULT = [];
```
cameronwaterman: `const result = [];`

src/datasources/system/SystemDataSource.ts
```diff
@@ -32,7 +33,8 @@ export class SystemDataSource extends DataSourceApi<SystemQuery> {
           ],
         });
       } else {
-        throw Error("Not implemented");
+        let metadataResponse = await getBackendSrv().post<{ data: SystemMetadata[] }>(this.baseUrl + '/query-systems', { projection: "new(id, alias, connected.data.state, grains.data.minion_blackout as locked, grains.data.boottime as systemStartTime, grains.data.productname as model, grains.data.manufacturer as vendor, grains.data.osfullname as osFullName, grains.data.ip4_interfaces as ip4Interfaces, grains.data.ip6_interfaces as ip6Interfaces, workspace)" });
```
cameronwaterman, line 36: Is there a better way we can construct this JSON string to help with readability?

cameronwaterman (approved):

release.config.js
```diff
@@ -16,4 +16,4 @@ module.exports = {
         }],
         '@semantic-release/git'
     ]
-}
```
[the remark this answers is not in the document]
cameronwaterman: It looks like that file isn't part of the changelist after merging in master anyways so I went ahead and resolved the comment

src/datasources/system/SystemDataSource.ts
ggracechoi: I don't think so? I tried putting each projection on a different line, but it gave me a bunch of errors/made the format weird. @mure do you have any suggestions?

mure: I think I'd extract this to a new file `constants.ts` and turn it into an array.
```ts
// constants.ts
export const defaultProjection = [
  'id',
  'alias',
  'connected.data.state'
]
```

and then add a method to the data source that transforms the array into the `"new(...)"` string.

mure: Also, we needed to add some logic to pull out the first ip address. Is that work tracked anywhere? We should add another user story or add it to the acceptance criteria of an existing one.

src/datasources/system/SystemDataSource.ts
```diff
@@ -19,6 +20,16 @@ export class SystemDataSource extends DataSourceApi<SystemQuery> {
...
 
+  async metricFindQuery(query: MyVariableQuery, options?: any) { // GC: or query: MyVariableQuery
+    // Retrieve DataQueryResponse based on query.
+    const response = await getBackendSrv().post<{ data: MyVariableQuery[] }>(this.baseUrl + '/query-systems', { projection: "new(id, alias)" });
+  
+    // Convert query results to a MetricFindValue[]
+    const values = response.data.map(frame => ({ text: frame.id }));
+  
+    return values;
+  }
```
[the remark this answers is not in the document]
ggracechoi: Oh no I accidentally pushed the variable stuff, so I just spent the last few minutes reverting the changes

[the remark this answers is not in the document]
mure: FWIW, I noticed that you've been working out of the same branch for every PR. It may be easier to create separate branches for each change, copying pieces out of the `users/grchoi/systems-datasource` branch as needed.

PR #9, opened 2023-07-26
test: Add test fixtures & implement `testDatasource` for tag

mure:
# Pull Request

## 🤨 Rationale

Writing unit tests for data sources is a little tricky because requests are made by the `BackendSrv` instance retrieved from a `getBackendSrv()` call. This PR adds some fixtures to make test specs cleaner and implements the tag's `testDatasource` method as an example.

## 👩‍💻 Implementation

I really wanted to use [Mock Service Worker](https://mswjs.io/), so that we could mock routes at the network level, but the `BackendSrv` implementation is apart of [Grafana's unpackaged code](https://github.com/grafana/grafana/blob/main/public/app/core/services/backend_srv.ts), so we have to mock it out.

Previously, we were using `jest.mock` to monkey patch the [@grafana/runtime module](https://github.com/ni/systemlink-grafana-plugins/blob/89eca48b245cac26b7c2934807e974f87f4b8294/src/datasources/notebook/datasource.test.ts#L13), but it only applies to the file that called it, so it's difficult to refactor. Instead, we can just inject the `BackendSrv` in DataSource's constructors. Then [jest-mock-extended](https://github.com/marchaos/jest-mock-extended) does a lot of the heavy lifting, letting us create a type-safe mock of the `BackendSrv` interface. The mock even comes with a `calledWith()` extension, so we can verify that the correct route is called.

## 🧪 Testing

This is self-explanatory 😄

## ✅ Checklist


package.json
```diff
@@ -8,6 +8,7 @@
     "dev": "webpack -w -c ./.config/webpack/webpack.config.ts --env development",
     "test": "jest --watch --onlyChanged",
     "test:ci": "jest --passWithNoTests --maxWorkers 4",
+    "test:template": "npx plop datasource Test && npx tsc --noEmit && rimraf src/datasources/test",
```
mure, line 11: I added this script to make it easy to test changes to the data source template. It generates a new data source, sees if it compiles, and then deletes it.

cameronwaterman (approved):

PR #10, opened 2023-07-29
test: fixtures for unit testing QueryEditors

mure:
# Pull Request

## 🤨 Rationale

We need to be able to easily write automated tests for the UI (query editor) of a data source plugin.

## 👩‍💻 Implementation

- Adds `renderQueryEditor` method to simplify writing unit tests with [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/)
- Adds a wrapper for the `InlineField` component that automatically assigns an id to the form control. This lets us use the `getByLabelText` selector in tests (see: [HTML attribute for](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/for))
- Implemented the tag path input in the `TagQueryEditor` component so I had something to write tests against

## 🧪 Testing

This change is mostly adding tests and test helpers.

## ✅ Checklist


src/datasources/tag/TagDataSource.ts
```diff
@@ -32,14 +32,20 @@ export class TagDataSource extends DataSourceApi<TagQuery> {
         refId: target.refId,
         fields: [
           { name: 'Time', values: [from, to], type: FieldType.time },
-          { name: 'Value', values: [target.constant, target.constant], type: FieldType.number },
+          { name: 'Value', values: [2.72, 3.14], type: FieldType.number },
```
cameronwaterman, line 35: Can we pull these values out to constants?

cameronwaterman (approved):

src/datasources/tag/TagDataSource.ts
mure: These aren't meaningful values - they'll be removed in the next change. (it's _e_ and π 😅)

PR #11, opened 2023-08-02
feat: implement metricFindQuery 

ggracechoi:
# Pull Request

## 🤨 Rationale

When users create a query variable, the dropdown list needs to populate with all the systems the user has access to. The user should also be able to look up a system by typing in the system alias. This PR satisfies [this user story](https://dev.azure.com/ni/DevCentral/_workitems/edit/2453816).

## 👩‍💻 Implementation

The _metricFindQuery_ method was implemented using the _/query-systems_ route. Two projections were made to extract the _id_ and the _alias_. Additionally, a new _VariableQuery_ interface was created in _types.ts_. 

## 🧪 Testing

The web dev tools were used to test this change. We checked to see if the query variable populated with all the systems that were present in the response.

## ✅ Checklist

- [ X ] This PR has a title that follows the [commit message format](https://github.com/ni/systemlink-grafana-plugins#commit-message-format).


cameronwaterman (changes requested):
@ggracechoi - It looks like your branch contains merge conflicts with the main branch. Please resolve those first and then reset my review.

src/datasources/system/components/SystemQueryEditor.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { InlineField, InlineFieldRow, RadioButtonGroup } from '@grafana/ui';
+import { InlineField, InlineFieldRow, Input, RadioButtonGroup } from '@grafana/ui';
```
cameronwaterman, line 2: Do we need this import?

cameronwaterman (approved):
We'll need to add tests for this stuff at some point but for the sake of time we can defer them for now

src/datasources/system/SystemDataSource.ts
```diff
@@ -40,6 +40,13 @@ export class SystemDataSource extends DataSourceApi<SystemQuery> {
     return NetworkUtils.getIpAddressFromInterfaces(ip4Interface) || NetworkUtils.getIpAddressFromInterfaces(ip6Interface);
   }
 
+  async metricFindQuery(query: VariableQuery, options?: any) {
```
cameronwaterman, line 43: This function needs a return type annotation

cameronwaterman: Also, most of the example usage I found of `metricFindQuery` returned `Promise<MetricFindValue>`, where this currently returns simple `MetricFindValue`. Are you sure this works as intended?

cameronwaterman (commented):
Changing my review until open questions are resolved

cameronwaterman (changes requested):
Changing my review until open questions are resolved

src/datasources/system/SystemDataSource.ts
```diff
@@ -53,20 +60,24 @@ export class SystemDataSource extends DataSourceApi<SystemQuery> {
           ],
         });
       } else {
-        let metadataResponse = await getBackendSrv().post<{ data: SystemMetadata[] }>(this.baseUrl + '/query-systems', { projection: this.transformProjection(defaultProjection) });
-        
+        const resolvedId = getTemplateSrv().replace(target.systemId);
```
mure, line 63:
```suggestion
        const resolvedId = getTemplateSrv().replace(target.systemId, options.scopedVars);
```

I'm not exactly sure when this is necessary, but it's part of the data source examples, so we should follow it.

src/datasources/system/components/SystemQueryEditor.tsx
```diff
@@ -13,13 +13,26 @@ export function SystemQueryEditor({ query, onChange, onRunQuery }: Props) {
...
   return (
     <div>
       <InlineFieldRow >
         <InlineField label="Query type">
           <RadioButtonGroup options={enumToOptions(QueryType)} onChange={onQueryTypeChange} value={query.queryKind} />
         </InlineField>
       </InlineFieldRow>
+      <InlineFieldRow>
+        <InlineField label="Minion ID">
+          <Input placeholder="All systems" onBlur={onIdBlur} />
```
mure, line 33: We need to set the `defaultValue` prop here to the value in the query so it's filled in when the user re-opens the query editor.

src/datasources/system/module.ts
```diff
@@ -5,4 +5,5 @@ import { HttpConfigEditor } from 'core/components/HttpConfigEditor';
 
 export const plugin = new DataSourcePlugin(SystemDataSource)
   .setConfigEditor(HttpConfigEditor)
-  .setQueryEditor(SystemQueryEditor);
+  .setQueryEditor(SystemQueryEditor)
+  .setVariableQueryEditor(() => null);
```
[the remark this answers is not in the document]
mure: I told Grace to add this. Without it, Grafana renders an arbitrary "Query" text box that is required to be filled in when creating a variable.

src/datasources/system/components/SystemQueryEditor.tsx
```diff
@@ -13,13 +13,26 @@ export function SystemQueryEditor({ query, onChange, onRunQuery }: Props) {
...
+
   return (
     <div>
       <InlineFieldRow >
         <InlineField label="Query type">
           <RadioButtonGroup options={enumToOptions(QueryType)} onChange={onQueryTypeChange} value={query.queryKind} />
         </InlineField>
       </InlineFieldRow>
+      <InlineFieldRow>
+        <InlineField label="Minion ID">
```
mure, line 32: If this field can be either the alias or the id, we should rename its label to something more generic, like "System name". We could also add the `tooltip` prop with a description of the accepted values.

src/datasources/system/types.ts
```diff
@@ -6,7 +6,8 @@ export enum QueryType {
 }
 
 export interface SystemQuery extends DataQuery {
-  queryKind: QueryType
+  queryKind: QueryType,
+  systemId: string
```
mure, line 10: Similar to my other comment, since this field can be either the alias or the id, we should name it more generically.

src/datasources/system/components/SystemQueryEditor.tsx
cameronwaterman: We discussed this with Mark as a "nice-to-have" in case they accidentally enter the alias instead of the id (when not using a query variable). Maybe it doesn't matter that much, but it might be confusing to users as "name" doesn't map to anything in the API. If we change the label I second adding some context help that guides the user.

mure: Yeah that's a fair point. We do use "name" in the Systems UI, though:

image

I could go either way. @prestwick any input?

src/datasources/system/SystemDataSource.ts
```diff
@@ -40,6 +41,13 @@ export class SystemDataSource extends DataSourceApi<SystemQuery> {
     return NetworkUtils.getIpAddressFromInterfaces(ip4Interface) || NetworkUtils.getIpAddressFromInterfaces(ip6Interface);
   }
 
+  async metricFindQuery(query: VariableQuery, options?: any): Promise<MetricFindValue[]> {
+    const response = await getBackendSrv().post<{ data: VariableQuery[] }>(this.baseUrl + '/query-systems', { projection: "new(id, alias)" });
+    const values = response.data.map(frame => ({ text: frame.alias, value: frame.id }));
+
+    return values;
```
cameronwaterman, line 48: The return type annotation states `Promise<MetricFindValue[]>`, but I don't think that is the type of this object?

mure (approved):

src/datasources/system/SystemDataSource.ts
```diff
@@ -53,20 +73,27 @@ export class SystemDataSource extends DataSourceApi<SystemQuery> {
...
-        
+        const resolvedId = getTemplateSrv().replace(target.systemName, options.scopedVars);
+        const postBody = {
+          filter: resolvedId ? `id = "${resolvedId}" || alias = "${resolvedId}"` : '',
+          projection: this.transformProjection(defaultProjection),
+          orderBy: 'createdTimeStamp DESC'
+        };
+        let metadataResponse = await getBackendSrv().post<{ data: SystemMetadata[] }>(this.baseUrl + '/query-systems', postBody);
+        let aliasNameResponse = await getBackendSrv().get< AuthResponse >(this.aliasUrl + '/auth', postBody);
+        console.log(aliasNameResponse);
```
cameronwaterman, line 84: Remove this

cameronwaterman (approved):

PR #14, opened 2023-08-11
Add launch configuration for VS Code

mure:
# Pull Request

## 🤨 Rationale

VS Code can attach to an instance of Chrome, which allows you to set and debug breakpoints right in the editor. This is a better workflow then having to open the dev tools separately in Chrome.

## 👩‍💻 Implementation

- Added a "task" that starts up Docker compose and `npm run dev`.
- Added a launch configuration that runs the task and then attaches to a new Chrome window.

## 🧪 Testing

Tried a few scenarios:
- Running the debugger and hitting a breakpoint
- Launching when Docker engine isn't running - correctly throws an error
- Running when the Docker compose is already started - restarts if needed

## ✅ Checklist

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,20 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "type": "chrome",
+      "request": "launch",
+      "name": "Launch Grafana in Chrome",
+      "url": "http://localhost:3000",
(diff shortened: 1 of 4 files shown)


cameronwaterman (approved):
Could we add a short blurb about this to the README?

mure:
> Could we add a short blurb about this to the README?

Done

PR #27, opened 2023-09-11
fix(tag)!: use tag name instead of 'value' for field name

mure:
# Pull Request

## 🤨 Rationale

Before this change, the data source would return a data frame with a name property set to the tag's name. The historical/current field was always named 'value':

**name: my.tag**
| time                | value |
|---------------------|-------|
| 2020-01-02 03:04:00 | 45.0  |
| 2020-01-02 03:05:00 | 47.0  |
| 2020-01-02 03:06:00 | 48.0  |

With this change, the 'value' field will take on the tag's name instead:

| time                | my.tag |
|---------------------|-------|
| 2020-01-02 03:04:00 | 45.0  |
| 2020-01-02 03:05:00 | 47.0  |
| 2020-01-02 03:06:00 | 48.0  |

Grafana's documentation doesn't provide any recommendations on which format is preferable, but the new approach makes the explore mode more usable for visualizing multiple tags.

Fixes [this bug](https://dev.azure.com/ni/DevCentral/_workitems/edit/2504138).

## 👩‍💻 Implementation

- Use the tag's name as the field name instead of the frame name

## 🧪 Testing

- Manually tested explore mode and dashboard mode to make sure that any core workflows haven't broken

## ✅ Checklist

**Unfortunately, I think we need to consider this a breaking change. It's possible for a dashboard to be configured in a way that relies on the fields being named 'value'. I know [this dashboard](https://dev.lifecyclesolutions.ni.com/dashboards/d/b73d4191-1060-4648-8649-8ad48142a855/canvas-demo?orgId=1) that I helped Josh set up will be broken.**

### DIFF
--- a/src/datasources/tag/TagDataSource.test.ts
+++ b/src/datasources/tag/TagDataSource.test.ts
@@ -38,8 +38,7 @@ describe('queries', () => {
 
     expect(result.data).toEqual([
       {
-        fields: [{ name: 'value', values: ['3.14'] }],
-        name: 'my.tag',
+        fields: [{ name: 'my.tag', values: ['3.14'] }],
         refId: 'A',
       },
     ]);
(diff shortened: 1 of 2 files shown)


cameronwaterman (approved):
Where did the request for this change come from? If we don't already, let's make sure we have a work item tracking this.

mure:
@cameronwaterman the origin is this bug: https://dev.azure.com/ni/DevCentral/_workitems/edit/2504138

PR #38, opened 2023-09-26
feat(workspace): Workspace data source returns workspaces

cameronwaterman:
# Pull Request

## 🤨 Rationale

https://dev.azure.com/ni/DevCentral/_workitems/edit/2527388

## 👩‍💻 Implementation

Implement `runQuery` and related functionality.

## 🧪 Testing

- Manually validated functionality
- Added new auto tests

## ✅ Checklist

### DIFF
--- a/src/datasources/workspace/WorkspaceDataSource.test.ts
+++ b/src/datasources/workspace/WorkspaceDataSource.test.ts
@@ -15,7 +15,7 @@ beforeEach(() => {
   [ds, backendSrv] = setupDataSource(WorkspaceDataSource);
 });
 
-fdescribe('testDatasource', () => {
+describe('testDatasource', () => {
   test('returns success', async () => {
     backendSrv.fetch
       .calledWith(requestMatching({ url: '/niuser/v1/workspaces' }))
--- a/src/datasources/workspace/WorkspaceDataSource.test.ts
+++ b/src/datasources/workspace/WorkspaceDataSource.test.ts
@@ -4,14 +4,12 @@ import _ from 'lodash';
 import {
   createFetchError,
   createFetchResponse,
-  defaultQueryOptions,
   getQueryBuilder,
   mockTimers,
   requestMatching,
   setupDataSource,
 } from 'test/fixtures';
 import { WorkspaceDataSource } from './WorkspaceDataSource';
-import { ArrayVector } from '@grafana/data';
 import { WorkspaceQuery } from './types';
 
 let ds: WorkspaceDataSource, backendSrv: MockProxy<BackendSrv>;
--- a/src/datasources/workspace/WorkspaceDataSource.test.ts
+++ b/src/datasources/workspace/WorkspaceDataSource.test.ts
@@ -1,0 +1,36 @@
+import { BackendSrv } from '@grafana/runtime';
+import { MockProxy } from 'jest-mock-extended';
+import _ from 'lodash';
+import {
+  createFetchError,
+  createFetchResponse,
+  requestMatching,
+  setupDataSource,
+} from 'test/fixtures';
+import { WorkspaceDataSource } from './WorkspaceDataSource';
+
+let ds: WorkspaceDataSource, backendSrv: MockProxy<BackendSrv>;
+
+beforeEach(() => {
+  [ds, backendSrv] = setupDataSource(WorkspaceDataSource);
+});
+
+fdescribe('testDatasource', () => {
+  test('returns success', async () => {
+    backendSrv.fetch
+      .calledWith(requestMatching({ url: '/niuser/v1/workspaces' }))
+      .mockReturnValue(createFetchResponse(25));
+
+    const result = await ds.testDatasource();
+
+    expect(result.status).toEqual('success');
+  });
+
+  test('bubbles up exception', async () => {
+    backendSrv.fetch
+      .calledWith(requestMatching({ url: '/niuser/v1/workspaces' }))
+      .mockReturnValue(createFetchError(400));
+
+    await expect(ds.testDatasource()).rejects.toHaveProperty('status', 400);
+  });
+});
--- a/src/datasources/workspace/WorkspaceDataSource.test.ts
+++ b/src/datasources/workspace/WorkspaceDataSource.test.ts
@@ -48,19 +48,17 @@ describe('queries', () => {
     const result = await ds.query(buildQuery({}));
 
     expect(result.data[0]).toHaveProperty('fields', [
-      { config: {}, name: 'ID', type: 'number', values: new ArrayVector(['1', '2']) },
-      { config: {}, name: 'Name', type: 'string', values: new ArrayVector(['Default workspace', 'Other workspace']) }
+      { name: 'ID', values: ['1', '2'] },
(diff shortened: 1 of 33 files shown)


src/datasources/workspace/WorkspaceDataSource.ts
```diff
@@ -10,30 +12,38 @@ export class WorkspaceDataSource extends DataSourceBase<WorkspaceQuery> {
...
+  shouldRunQuery(): boolean {
     return true;
   }
 
   async testDatasource(): Promise<TestingStatus> {
     await this.get(this.baseUrl + '/workspaces');
     return { status: 'success', message: 'Data source connected and authentication successful!' };
   }
+
+  private workspaceToFields(app: string, workspaces: Workspace[]): Array<{ name: string, values: string[] }> {
```
cameronwaterman, line 36: I feel like there's a better way to write this function using the array API but my JS wheels are in need of some oil - maybe you'll know of a better implementation.

src/datasources/workspace/components/WorkspaceQueryEditor.tsx
```diff
@@ -1,33 +1,16 @@
...
   return (
-    <>
-      <InlineField label="Constant">
-        <Input onChange={onConstantChange} value={constant} width={8} type="number" step="0.1" />
-      </InlineField>
-      <InlineField label="Query Text" labelWidth={16} tooltip="Not used yet">
-        <Input onChange={onQueryTextChange} value={queryText || ''} />
-      </InlineField>
-    </>
+    <span>This data source does not include a query editor - it simply returns all of the SystemLink Workspaces.</span>
```
cameronwaterman, line 14: @MelissaHilliard - If you have a few could you review this string? It shows up in a panel after selecting this data source.

image

MelissaHilliard:
```suggestion
    <span>This data source returns all SystemLink workspaces and does not include a query editor.</span>
```

src/datasources/workspace/WorkspaceDataSource.ts
```diff
@@ -10,30 +12,38 @@ export class WorkspaceDataSource extends DataSourceBase<WorkspaceQuery> {
...
+    const ids: string[] = [];
+    const names: string[] = [];
+    const queryVariableEditor = app === 'dashboard';
```
mure, line 39: I don't think this is sufficient, unfortunately. When you're in the panel editor, the value is `panel-editor`, but when you're viewing a dashboard, both the panel queries and the variable queries are `dashboard`.

For a variable query, the request looks like:
```
{
    "app": "dashboard",
    "requestId": "deb976da-36a0-42ea-bd05-8b4efb8ec870",
    "timezone": "",
    "range": {
        "from": "2023-09-27T13:49:13.572Z",
        "to": "2023-09-27T19:49:13.572Z",
        "raw": {
            "from": "now-6h",
            "to": "now"
        }
    },
    "interval": "",
    "intervalMs": 0,
    "targets": [
        {
            "refId": "variable-query"
        }
    ],
    "scopedVars": {
        "variable": {}
    },
    "startTime": 1695844153572
}
```
So we may be able to key off the `refId` property...but I'm not sure if we can rely on that not to change at some point in the future.

mure (changes requested):

src/datasources/workspace/WorkspaceDataSource.ts
```diff
@@ -10,30 +12,38 @@ export class WorkspaceDataSource extends DataSourceBase<WorkspaceQuery> {
...
-      refId: query.refId,
-      fields: [
-        { name: 'Time', values: [range.from.valueOf(), range.to.valueOf()], type: FieldType.time },
-        { name: 'Value', values: [query.constant, query.constant], type: FieldType.number },
-      ],
-    };
+  defaultQuery = {};
+
+  async runQuery(_query: WorkspaceQuery, { app }: DataQueryRequest): Promise<DataFrame> {
+    return toDataFrame({ fields: this.workspaceToFields(app, await this.getWorkspaces()) });
```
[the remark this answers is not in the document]
mure: Oh, I started playing around with the code and I see that the variables don't seem to work unless this is a true DataFrame.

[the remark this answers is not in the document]
mure: Turns out it's a bug that's been fixed but not yet released:
https://github.com/grafana/grafana/commit/dc6675cade5236765a65b9e8f801c61296c50e22

src/datasources/workspace/WorkspaceDataSource.ts
```diff
@@ -14,26 +15,36 @@ export class WorkspaceDataSource extends DataSourceBase<WorkspaceQuery> {
...
+  private workspacesToFields(workspaces: Workspace[]): Array<{ name: string, values: string[] }> {
+    const ids: string[] = [];
+    const names: string[] = [];
+    workspaces.forEach((workspace: Workspace) => {
+      ids.push(workspace.id);
+      names.push(workspace.name);
+    });
+    return [
+      { name: 'ID', values: ids },
+      { name: 'Name', values: names }
```
mure, line 47: Probably not super important, but I think we've been consistent about using lowercase for field names:
https://github.com/ni/systemlink-grafana-plugins/blob/ef51b3aae45e502419d798ff3d910d4d9d9fc8f1/src/datasources/system/SystemDataSource.ts#L41-L53

src/datasources/workspace/WorkspaceDataSource.ts
```diff
@@ -14,26 +15,36 @@ export class WorkspaceDataSource extends DataSourceBase<WorkspaceQuery> {
...
-      refId: query.refId,
-      fields: [
-        { name: 'Time', values: [range.from.valueOf(), range.to.valueOf()], type: FieldType.time },
-        { name: 'Value', values: [query.constant, query.constant], type: FieldType.number },
-      ],
-    };
+  defaultQuery = {};
+
+  async runQuery(_query: WorkspaceQuery): Promise<DataFrameDTO> {
+    return { fields: this.workspacesToFields( await this.getWorkspaces()) };
```
mure, line 21: I think we had talked about having this just list workspace names instead of both the id and name. 
Admittedly, the usefulness of this data source in a panel is really low, so it probably doesn't matter, but I think if someone does use it they wouldn't care about the ids.

src/datasources/workspace/WorkspaceDataSource.test.ts
```diff
@@ -4,17 +4,23 @@ import _ from 'lodash';
...
+import { WorkspaceQuery } from './types';
 
 let ds: WorkspaceDataSource, backendSrv: MockProxy<BackendSrv>;
 
 beforeEach(() => {
   [ds, backendSrv] = setupDataSource(WorkspaceDataSource);
 });
 
+const buildQuery = getQueryBuilder<WorkspaceQuery>()({});
+mockTimers();
```
mure, line 22: I don't think this is necessary since `setTimeout()` won't be called from these tests. This was needed in the tag tests to be able to test the retry logic.

mure (approved):

PR #40, opened 2023-09-27
feat(data-frame): Data Frame data source displays associated Workspace

cameronwaterman:
# Pull Request

## 🤨 Rationale

https://dev.azure.com/ni/DevCentral/_workitems/edit/2527389

## 👩‍💻 Implementation

Show Data Table Workspace name as `description` text is Data Table selection drop-down and move the ID to `title` (hover text). This involved needing to migrate the Data Table data source to extend `DataSourceBase`.

## 🧪 Testing

Tests continue to pass.

## ✅ Checklist

### DIFF
--- a/src/core/DataSourceBase.ts
+++ b/src/core/DataSourceBase.ts
@@ -1,4 +1,5 @@
 import {
+  DataFrame,
   DataFrameDTO,
   DataQueryRequest,
   DataQueryResponse,
@@ -17,7 +18,7 @@ export abstract class DataSourceBase<TQuery extends DataQuery> extends DataSourc
   }
 
   abstract defaultQuery: Partial<TQuery> & Omit<TQuery, 'refId'>;
-  abstract runQuery(query: TQuery, options: DataQueryRequest): Promise<DataFrameDTO>;
+  abstract runQuery(query: TQuery, options: DataQueryRequest): Promise<DataFrame | DataFrameDTO>;
   abstract shouldRunQuery(query: TQuery): boolean;
   abstract testDatasource(): Promise<TestingStatus>;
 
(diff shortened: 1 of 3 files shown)


src/core/DataSourceBase.ts
```diff
@@ -46,8 +47,8 @@ export abstract class DataSourceBase<TQuery extends DataQuery> extends DataSourc
     }
   }
 
-  get<T>(url: string) {
-    return this.fetch<T>({ method: 'GET', url });
+  get<T>(url: string, params: Record<string, any> | undefined = undefined) {
```
mure, line 50: This is equivalent but shorter:
```suggestion
  get<T>(url: string, params?: Record<string, any>) {
```

src/datasources/data-frame/components/DataFrameQueryEditor.tsx
```diff
@@ -40,10 +41,15 @@ export const DataFrameQueryEditor = (props: Props) => {
   };
 
   const loadTableOptions = _.debounce((query: string, cb?: LoadOptionsCallback<string>) => {
-    datasource
+    const tablesPromise = datasource
```
mure, line 44:
```
Promise.all([datasource.queryTables(query), datasource.getWorkspaces()])
  .then(([tables, workspaces]) => cb?.(tables.map((t) => ({ label: t.name, value: t.id, title: t.id, description: getWorkspaceName(workspaces, t.workspace) }))))
  .catch(handleError);
```

mure: I think this works the same

mure (approved):
Thanks for doing the refactor to extend from DataSourceBase. Glad to see that we didn't have to modify the tests at all.

src/core/DataSourceBase.ts
cameronwaterman: I could've swore I had this initially and my editor was complaining about it. Seems to be working now.

src/datasources/data-frame/components/DataFrameQueryEditor.tsx
cameronwaterman: Much cleaner - thanks!

PR #41, opened 2023-10-04
feat(tag): Workspace variable resolution

cameronwaterman:
# Pull Request

## 🤨 Rationale

https://ni.visualstudio.com/DevCentral/_workitems/edit/2527392

## 👩‍💻 Implementation

Prepend dashboard variables to list of workspaces and attempt to resolve variables populated in the workspace field prior to querying.

## 🧪 Testing

- Validated functionality locally
- Add new test and existing tests continue to pass

## ✅ Checklist

### DIFF
--- a/src/datasources/tag/components/TagQueryEditor.test.tsx
+++ b/src/datasources/tag/components/TagQueryEditor.test.tsx
@@ -4,7 +4,7 @@ import { select } from 'react-select-event';
 import { setupRenderer } from 'test/fixtures';
 import { TagDataSource } from '../TagDataSource';
 import { TagQuery, TagQueryType } from '../types';
-import { TagQueryEditor } from './TagQueryEditor';;
+import { TagQueryEditor } from './TagQueryEditor';
 
 const render = setupRenderer(TagQueryEditor, TagDataSource);
 const workspacesLoaded = () => waitForElementToBeRemoved(screen.getByTestId('Spinner'));
--- a/src/datasources/tag/components/TagQueryEditor.test.tsx
+++ b/src/datasources/tag/components/TagQueryEditor.test.tsx
@@ -6,6 +6,13 @@ import { TagDataSource } from '../TagDataSource';
 import { TagQuery, TagQueryType } from '../types';
 import { TagQueryEditor } from './TagQueryEditor';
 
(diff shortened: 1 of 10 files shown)


src/datasources/tag/components/TagQueryEditor.tsx
```diff
@@ -28,6 +29,23 @@ export function TagQueryEditor({ query, onChange, onRunQuery, datasource }: Prop
...
+      return [];
+    }
+  }
+
+  const getVariableOptions = () => {
+    return getTemplateSrv()
+      .getVariables()
+      .map((variable) => ({ label: '$' + variable.name, value: '$' + variable.name }));
+  };
+
```
mure, line 48: I think it would be nice if we could add this logic to the `useWorkspaceOptions` hook, so we can reuse it. 
We should add `templateSrv` to the `DataSourceBase` constructor and then `useWorkspaceOptions` can call into it to get the variables. It'll be a bit cleaner too, because we won't need the `if (workspaces.loading)` check.

src/datasources/tag/components/TagQueryEditor.test.tsx
```diff
@@ -6,6 +6,13 @@ import { TagDataSource } from '../TagDataSource';
 import { TagQuery, TagQueryType } from '../types';
 import { TagQueryEditor } from './TagQueryEditor';
 
+jest.mock('@grafana/runtime', () => ({
+  ...jest.requireActual('@grafana/runtime'),
+  getTemplateSrv: () => ({ getVariables: getVariablesMock }),
+}));
+
+const getVariablesMock = jest.fn(() => [{ name: 'testVar' }]);
```
mure, line 14: If you implement the approach in my other comment, this won't work anymore. But it should actually be simpler. We can add the mock implementation for `getVariables` here:

https://github.com/ni/systemlink-grafana-plugins/blob/335fb5f56938ce33c34993df5056440e1a58a245/src/test/fixtures.ts#L21-L23

src/datasources/tag/components/TagQueryEditor.tsx
cameronwaterman: Done. This PR is ready for re-review.

mure (approved):

PR #42, opened 2023-10-04
feat(tag): include properties in query result

mure:
# Pull Request

## 🤨 Rationale

https://dev.azure.com/ni/DevCentral/_workitems/edit/2504597

## 👩‍💻 Implementation

Pretty straightforward - `TagQueryEditor` now has an `InlineSwitch` component that lets users toggle whether or not to show properties. In `TagDataSource`, we map the tag's properties to DataFrame fields and conditionally append them to the result. We ignore properties that start with "nitag" because they're used by the historian service and not useful to users. We also now return the tag's last updated timestamp for the "Current" query mode.

## 🧪 Testing

I decided to make the leap to using [snapshot testing](https://jestjs.io/docs/snapshot-testing) for some of the tag data source tests. I've been thinking about it for a while, but this change made updating the test assertions tedious enough that I tried it out. It's _really_ nice for testing the data source `query` method. Instead of manually typing out the expected result, Jest generates a snapshot file for us that stores the results. Then in the future if the snapshot changes, the test will fail.

## ✅ Checklist

### DIFF
--- a/src/datasources/tag/components/TagQueryEditor.tsx
+++ b/src/datasources/tag/components/TagQueryEditor.tsx
@@ -68,5 +68,5 @@ const tooltips = {
   workspace: `The workspace to search for the given tag path. If left blank, the plugin
               finds the most recently updated tag in any workspace.`,
 
-  properties: `If toggled, include the tag's properties in the query result.`,
+  properties: `If enabled, include the tag properties in the query result.`,
 };
--- a/src/datasources/tag/components/TagQueryEditor.tsx
+++ b/src/datasources/tag/components/TagQueryEditor.tsx
@@ -1,5 +1,5 @@
 import { QueryEditorProps, SelectableValue } from '@grafana/data';
-import { AutoSizeInput, RadioButtonGroup, Select } from '@grafana/ui';
+import { AutoSizeInput, InlineSwitch, RadioButtonGroup, Select } from '@grafana/ui';
 import { InlineField } from 'core/components/InlineField';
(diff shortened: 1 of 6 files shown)


src/datasources/tag/components/TagQueryEditor.tsx
```diff
@@ -59,4 +67,6 @@ const tooltips = {
 
   workspace: `The workspace to search for the given tag path. If left blank, the plugin
               finds the most recently updated tag in any workspace.`,
+
+  properties: `If toggled, include the tag's properties in the query result.`,
```
cameronwaterman, line 71: Please review this string @MelissaHilliard

src/datasources/tag/types.ts
```diff
@@ -9,14 +9,15 @@ export interface TagQuery extends DataQuery {
   type: TagQueryType;
   path: string;
   workspace: string;
+  properties: boolean;
 }
 
 export interface TagWithValue {
-  current: { value: { value: string } } | null;
+  current: { value: { value: string }; timestamp: string } | null;
```
cameronwaterman, line 16: Can we split this out onto multiple lines, readability isn't great.

cameronwaterman (approved):

src/datasources/tag/components/TagQueryEditor.tsx
MelissaHilliard:
```suggestion
  properties: `If enabled, include tag properties in the query result.`,
```

PR #43, opened 2023-10-05
feat(system): Filter system metadata by workspace and variable chaining

cameronwaterman:
# Pull Request

## 🤨 Rationale

https://ni.visualstudio.com/DevCentral/_workitems/edit/2527391
https://ni.visualstudio.com/DevCentral/_workitems/edit/2541892

## 👩‍💻 Implementation

- Add workspace field to metadata query which is capable of resolving variables.
- Add variable resolution logic to `metricFindQuery`.

## 🧪 Testing

Added tests and existing tests continue to pass.

## ✅ Checklist

### DIFF
--- a/src/core/utils.ts
+++ b/src/core/utils.ts
@@ -37,8 +37,17 @@ export function throwIfNullish<T>(value: T, error: string | Error): NonNullable<
 export function useWorkspaceOptions<DSType extends DataSourceBase<any>>(datasource: DSType) {
   return useAsync(async () => {
     const workspaces = await datasource.getWorkspaces();
-    return workspaces.map(w => ({ label: w.name, value: w.id }));
+    const workspaceOptions = workspaces.map(w => ({ label: w.name, value: w.id }));
+    workspaceOptions?.unshift(...getVariableOptions(datasource))
+    return workspaceOptions;
   });
+}
+
+/** Gets Dashboard variables as an array of {@link SelectableValue}. */
+export function getVariableOptions<DSType extends DataSourceBase<any>>(datasource: DSType) {
+  return datasource.templateSrv
+    .getVariables()
+    .map((variable) => ({ label: '$' + variable.name, value: '$' + variable.name }));
 }
 
 export function getWorkspaceName(workspaces: Workspace[], id: string) {
--- a/src/core/utils.ts
+++ b/src/core/utils.ts
@@ -47,7 +47,8 @@ export function useWorkspaceOptions<DSType extends DataSourceBase<any>>(datasour
 export function getVariableOptions<DSType extends DataSourceBase<any>>(datasource: DSType) {
   return datasource.templateSrv
     .getVariables()
-    .map((variable) => ({ label: '$' + variable.name, value: '$' + variable.name }));
+    .filter(variable => (variable as any).current.value)
+    .map(variable => ({ label: '$' + variable.name, value: '$' + variable.name }));
 }
 
 export function getWorkspaceName(workspaces: Workspace[], id: string) {
--- a/src/core/utils.ts
+++ b/src/core/utils.ts
@@ -47,7 +47,7 @@ export function useWorkspaceOptions<DSType extends DataSourceBase<any>>(datasour
 export function getVariableOptions<DSType extends DataSourceBase<any>>(datasource: DSType) {
   return datasource.templateSrv
     .getVariables()
-    .filter(variable => (variable as any).current.value)
+    .filter((variable: any) => !variable.datasource || variable.datasource.uid !== datasource.uid)
     .map(variable => ({ label: '$' + variable.name, value: '$' + variable.name }));
 }
 
--- a/src/datasources/system/SystemDataSource.test.ts
+++ b/src/datasources/system/SystemDataSource.test.ts
@@ -10,7 +10,7 @@ beforeEach(() => {
   [ds, backendSrv, templateSrv] = setupDataSource(SystemDataSource);
 });
 
-const buildQuery = getQueryBuilder<SystemQuery>()({ systemName: '' });
+const buildQuery = getQueryBuilder<SystemQuery>()({ systemName: '', workspace: '' });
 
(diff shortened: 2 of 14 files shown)


src/datasources/system/components/SystemQueryEditor.tsx
```diff
@@ -49,4 +69,5 @@ const tooltips = {
               Summary allows you to visualize the number of disconnected and connected systems.`,
   system: `Query for a specific system by its name or ID. If left blank, the plugin returns all
             available systems. You can enter a variable into this field.`,
+  workspace: `The workspace to search for the given system.`,
```
cameronwaterman, line 72: @MelissaHilliard - Please review this string when you get a second.

MelissaHilliard:
```suggestion
  workspace: `The workspace to search for the system specified.`,
```
We usually say "specified" but it's not a big deal either way

src/datasources/system/SystemDataSource.ts
```diff
@@ -33,7 +34,11 @@ export class SystemDataSource extends DataSourceBase<SystemQuery> {
         ],
       };
     } else {
-      const metadata = await this.getSystemMetadata(this.templateSrv.replace(query.systemName, options.scopedVars));
+      const metadata = await this.getSystemMetadata(
+        this.templateSrv.replace(query.systemName, options.scopedVars),
+        defaultProjection,
+        this.templateSrv.replace(query.workspace, options.scopedVars)
+      );
```
mure, line 41: I'm getting an error if I put in both a system and a workspace in the panel query editor. The generated filter is:

`id = "asdf" || alias = "asdf" workspace = "e2897cf7-6332-433c-8284-65b7b628f3f6"`

That's my fault - I think the join in `getSystemMetadata` needs to use `&&` as the separator. Although now that I think about it, it's not much of a use case to filter by both a specific system and a workspace. I guess you could have systems with the same name in different workspaces?

src/datasources/system/SystemDataSource.test.ts
```diff
@@ -114,6 +114,28 @@ test('queries for system variable values - single workspace', async () => {
   expect(backendSrv.fetch.mock.lastCall?.[0].data).toHaveProperty('filter', 'workspace = "1"');
 });
 
+test('attempts to replace variables in metadata query', async () => {
+  const workspaceVariable = '$workspace';
+  backendSrv.fetch.mockReturnValue(createFetchResponse({ data: fakeSystems.map(({ id, alias }) => ({ id, alias })) }));
```
mure, line 119:
```suggestion
  backendSrv.fetch.mockReturnValue(createFetchResponse({ data: fakeSystems }));
```

It doesn't really matter in this test, but this is the more correct response for the metadata query.

src/datasources/system/SystemDataSource.ts
```diff
@@ -70,7 +75,7 @@ export class SystemDataSource extends DataSourceBase<SystemQuery> {
   }
 
   async metricFindQuery({ workspace }: SystemVariableQuery): Promise<MetricFindValue[]> {
-    const metadata = await this.getSystemMetadata('', ['id', 'alias'], workspace);
+    const metadata = await this.getSystemMetadata('', ['id', 'alias'], this.templateSrv.replace(workspace));
```
mure, line 78: I played around with this locally. It's cool that the chained variable just works without any extra configuration.

mure (commented):
Maybe we could add a test to `SystemVariableQueryEditor.test.tsx` that tests if a user can select a variable option.

mure:
image

It would be nice if we could not add the variable you're currently editing to the dropdown, but I'm not sure how to know that.

src/datasources/system/SystemDataSource.ts
joshuaprewitt: I agree that filtering by system and workspace isn't really needed. I'm pretty sure that minionIDs have to be unique at least at the salt level. The workspace filter is mainly needed when viewing the metadata for all systems as well as for the system variable to limit the systems in the drop-down.

cameronwaterman: Okay, I'll look into only showing the workspace drop-down when "All systems" is selected.

cameronwaterman:
> image
> 
> It would be nice if we could not add the variable you're currently editing to the dropdown, but I'm not sure how to know that.

Yeah I thought this as well, but wasn't having much luck actually getting the current variable name in the context of the variable query editor. I'll spend a bit more time looking into this today.

src/core/utils.ts
```diff
@@ -47,7 +47,8 @@ export function useWorkspaceOptions<DSType extends DataSourceBase<any>>(datasour
 export function getVariableOptions<DSType extends DataSourceBase<any>>(datasource: DSType) {
   return datasource.templateSrv
     .getVariables()
-    .map((variable) => ({ label: '$' + variable.name, value: '$' + variable.name }));
+    .filter(variable => (variable as any).current.value)
```
cameronwaterman, line 50: @mure - Here is my proposed solution to not showing the newly-created variable in the drop-down. It's hacky, but seems to work fairly well in my testing. Let me know what you think.

mure: Looks like query variables have a `datasource` property:

image

Maybe we could hide variables that are from the same data source the user is currently creating a variable for?

I'm also fine with not doing anything. It's not really a big deal that users see an extra autocomplete option.

src/test/fixtures.ts
```diff
@@ -18,9 +18,26 @@ export function setupDataSource<T>(
       },
     }
   );
+  const mockVariables: TypedVariableModel[] = [{
```
mure, line 21: nit: I'd prefer to have this constant outside of the function for readability

mure (approved):

src/core/utils.ts
cameronwaterman: I've gone with checking the data source, it is cleaner

PR #46, opened 2023-10-18
fix(data-frame): Boolean and "Value" columns

cameronwaterman:
# Pull Request

## 🤨 Rationale

https://ni.visualstudio.com/DevCentral/_workitems/edit/2553637
https://ni.visualstudio.com/DevCentral/_workitems/edit/2553643

## 👩‍💻 Implementation

- For the boolean column issue, we're now converting the boolean stored in a string to it's proper type before converting to a data frame and transforming the rest of the values.
-  For the "Value" column issue, we're now just simply setting the column display name configuration to the column name.

## 🧪 Testing

- Verified fixes locally
- Updated existing test

## ✅ Checklist

### DIFF
--- a/src/datasources/data-frame/DataFrameDataSource.ts
+++ b/src/datasources/data-frame/DataFrameDataSource.ts
@@ -155,7 +155,9 @@ export class DataFrameDataSource extends DataSourceBase<DataFrameQuery> {
     this.transformBooleanFields(tableData, columns);
     const frame = toDataFrame(tableData);
     frame.fields.forEach(field => {
-      field.config.displayName = field.name;
+      if (field.name.toLocaleLowerCase() === 'value') {
+        field.config.displayName = field.name;
+      }
     })
     const transformer = standardTransformers.convertFieldTypeTransformer.transformer;
     const conversions = columns.map(({ name, dataType }) => ({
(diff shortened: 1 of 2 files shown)


src/datasources/data-frame/DataFrameDataSource.ts
```diff
@@ -152,7 +151,12 @@ export class DataFrameDataSource extends DataSourceBase<DataFrameQuery> {
     }
   }
 
-  private convertDataFrameFields(frame: DataFrame, columns: Column[]): DataFrame {
+  private convertDataFrameFields(tableData: TableData, columns: Column[]): DataFrame {
+    this.transformBooleanFields(tableData, columns);
+    const frame = toDataFrame(tableData);
+    frame.fields.forEach(field => {
+      field.config.displayName = field.name;
+    })
```
mure, line 159: I think we should only specify this when the column's name is 'value', otherwise it could technically be a breaking change. Some panels such as the graph let you set a display name, but it doesn't work when the field's configuration is set programmatically like this. The workaround is pretty simple and I'm not sure anyone will care, but my preference would be to scope the change smaller.

image

mure (approved):

PR #48, opened 2023-10-23
fix(tag): support `datatype` and `type` properties

mure:
# Pull Request

## 🤨 Rationale

The tag service had a [breaking change]( https://dev.azure.com/ni/DevCentral/_git/Skyline/commit/b3d9e6c8392b2d6f1b95de3512faf50d91cd410f?refName=refs%2Fheads%2Fmaster) that renamed the `datatype` field to `type`. We should handle both versions of the tag service to avoid any deployment issues.

## 👩‍💻 Implementation

- Added a separate interface for the legacy field names
- Check both `type` and `datatype`

## 🧪 Testing

- Added unit tests for the legacy fields

## ✅ Checklist

### DIFF
--- a/src/datasources/tag/types.ts
+++ b/src/datasources/tag/types.ts
@@ -12,19 +12,36 @@ export interface TagQuery extends DataQuery {
   properties: boolean;
 }
 
-export interface TagWithValue {
+interface TagWithValueBase {
   current: {
     value: { value: string };
     timestamp: string;
   } | null;
   tag: {
-    datatype: string;
     path: string;
     properties: Record<string, string> | null;
-    workspace: string;
-    workspace_id: string;
   };
 }
+
+// Legacy tag properties from SystemLink Server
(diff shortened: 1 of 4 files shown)


src/datasources/tag/types.ts
```diff
@@ -12,20 +12,37 @@ export interface TagQuery extends DataQuery {
...
   tag: {
-    datatype: string;
     path: string;
     properties: Record<string, string> | null;
-    workspace: string;
-    workspace_id: string;
   };
 }
 
+// Legacy tag properties from SystemLink Server
```
cameronwaterman, line 26: Is this comment accurate? It was my understanding that they changed to camelCase to match SLS.

mure: Shoot you're right. For some reason I was under the impression that it was the other way around. Probably wasn't worth writing the tests then.

cameronwaterman (approved):

PR #49, opened 2023-10-31
feat(dataframe): visualize table properties

mure:
# Pull Request

## 🤨 Rationale

https://dev.azure.com/ni/DevCentral/_workitems/edit/2566752

## 👩‍💻 Implementation

- Added a new field to the data frame query model, `type`. "Data" is the original behavior of the plugin and "Metadata" is a new mode that returns a table's properties as a data frame with two fields: name and value.
- I did some refactoring while prototyping. The major things were:
  - Cleaned up the data conversion logic to not rely on the transform built in to Grafana (it didn't work for booleans and had an awkward API)
  - Added logic to the fake routes in the tests to better match what the actual service does. We were returning all of the columns of fake data in every test, which wasn't technically correct.

## 🧪 Testing

- Added a unit test for the metadata query

## ✅ Checklist

### DIFF
--- a/src/datasources/data-frame/DataFrameDataSource.ts
+++ b/src/datasources/data-frame/DataFrameDataSource.ts
@@ -1,22 +1,7 @@
 import TTLCache from '@isaacs/ttlcache';
 import deepEqual from 'fast-deep-equal';
-import {
-  DataQueryRequest,
-  DataSourceInstanceSettings,
-  toDataFrame,
-  TableData,
-  FieldType,
-  standardTransformers,
-  DataFrame,
-  TimeRange
-} from '@grafana/data';
-import {
-  BackendSrv,
-  TemplateSrv,
-  TestingStatus,
-  getBackendSrv,
-  getTemplateSrv
-} from '@grafana/runtime';
+import { DataQueryRequest, DataSourceInstanceSettings, FieldType, TimeRange, FieldDTO, dateTime } from '@grafana/data';
+import { BackendSrv, TemplateSrv, TestingStatus, getBackendSrv, getTemplateSrv } from '@grafana/runtime';
 import {
   ColumnDataType,
   DataFrameQuery,
@@ -27,6 +12,7 @@ import {
   Column,
   defaultQuery,
   ValidDataFrameQuery,
+  DataFrameQueryType,
 } from './types';
 import { metadataCacheTTL } from './constants';
 import _ from 'lodash';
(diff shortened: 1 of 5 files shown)


src/datasources/data-frame/DataFrameDataSource.ts
```diff
@@ -45,32 +31,39 @@ export class DataFrameDataSource extends DataSourceBase<DataFrameQuery> {
 
   baseUrl = this.instanceSettings.url + '/nidataframe/v1';
 
-  defaultQuery = defaultQuery
+  defaultQuery = defaultQuery;
 
-  async runQuery(query: DataFrameQuery, { range, scopedVars, maxDataPoints }: DataQueryRequest): Promise<DataFrame> {
+  async runQuery(query: DataFrameQuery, { range, scopedVars, maxDataPoints }: DataQueryRequest) {
```
cameronwaterman, line 36: Are we purposely not specifying a return type here?

src/datasources/data-frame/components/DataFrameQueryEditor.tsx
```diff
@@ -109,5 +134,19 @@ export const DataFrameQueryEditor = (props: Props) => {
 const getVariableOptions = () => {
   return getTemplateSrv()
     .getVariables()
-    .map((v) => toOption('$' + v.name));
+    .map(v => toOption('$' + v.name));
+};
+
+const tooltips = {
```
cameronwaterman, line 140: @MelissaHilliard - Please review these strings when you get a chance.

cameronwaterman (approved):

src/datasources/data-frame/DataFrameDataSource.ts
mure: Oh, that wasn't intentional. I had removed it when I was prototyping different result formats.

src/datasources/data-frame/components/DataFrameQueryEditor.tsx
```diff
@@ -109,5 +134,19 @@ export const DataFrameQueryEditor = (props: Props) => {
...
+  queryType: `Data allows you to visualize the rows of data in a table. Metadata allows you
```
MelissaHilliard, line 141:
```suggestion
  queryType: `Specifies whether to visualize the data rows or properties associated with a table.

File: src/datasources/data-frame/components/DataFrameQueryEditor.tsx [hunk 4]
```diff
@@ -109,5 +134,19 @@ export const DataFrameQueryEditor = (props: Props) => {
...
+              to visualize the properties associated with a table.`,
```
MelissaHilliard, line 142:
```suggestion
`,
```

src/datasources/data-frame/components/DataFrameQueryEditor.tsx
```diff
@@ -109,5 +134,19 @@ export const DataFrameQueryEditor = (props: Props) => {
...
+
+  columns: `Specifies the columns to include in the response data.`,
+
+  decimation: `Specifies the method used to decimate the data.`,
+
+  filterNulls: `Filters out null and NaN values before decimating the data.`,
```
MelissaHilliard, line 148:
```suggestion
  filterNulls: `Specifies whether to filter out null and NaN values before decimating the data.`,
```

src/datasources/data-frame/components/DataFrameQueryEditor.tsx
```diff
@@ -109,5 +134,19 @@ export const DataFrameQueryEditor = (props: Props) => {
...
+
+  useTimeRange: `Queries only for data within the dashboard time range if the table index is a
```
MelissaHilliard, line 150:
```suggestion
  useTimeRange: `Specifies whether to query only for data within the dashboard time range if the table index is a
```

PR #50, opened 2023-11-01
feat(dataframe): Column variable support

cameronwaterman:
# Pull Request

## 🤨 Rationale

https://ni.visualstudio.com/DevCentral/_workitems/edit/2566754

## 👩‍💻 Implementation

- Added variable support to query editor columns field.
- Added variable query editor to data frame data source that displays a table id field. This field also supports variables. Implemented `metricFindQuery` to return the selected tables columns.
- Since the data frame query editor and variable query editor shared the majority of their logic, this was split out to a helper class that is accessed by both.

## 🧪 Testing

- Verified functionality locally
- Added new tests and existing tests continue to pass

## ✅ Checklist

### DIFF
--- a/src/core/utils.ts
+++ b/src/core/utils.ts
@@ -2,6 +2,7 @@ import { SelectableValue } from '@grafana/data';
 import { useAsync } from 'react-use';
 import { DataSourceBase } from './DataSourceBase';
 import { Workspace } from './types';
+import { TemplateSrv } from '@grafana/runtime';
 
 export function enumToOptions<T>(stringEnum: { [name: string]: T }): Array<SelectableValue<T>> {
   const RESULT = [];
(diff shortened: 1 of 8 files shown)


cameronwaterman:
@mure - This should be good to go again. The merge was pretty nasty so make sure you give it another look.

src/core/utils.ts
```diff
@@ -62,3 +63,22 @@ export function getWorkspaceName(workspaces: Workspace[], id: string) {
...
+/**
+ * Replace variables in an array of values.
+ * Useful for multi-value variables.
+ */
+export function replaceVariables(values: string[], templateSrv: TemplateSrv) {
+  const replaced: string[] = [];
+  values.forEach((col: string, index) => {
+    let value = col;
+    if (templateSrv.containsTemplate(col)) {
+      const variables = templateSrv.getVariables() as any[];
```
mure, line 76: Why do we need the `any[]` cast here?

src/datasources/data-frame/components/DataFrameQueryEditorCommon.tsx
```diff
@@ -0,0 +1,60 @@
...
+  readonly datasource: DataFrameDataSource;
+  readonly onChange: (value: DataFrameQuery) => void;
+  readonly query: ValidDataFrameQuery;
+  readonly onRunQuery: () => false | void;
+  readonly handleError: (error: Error) => void;
+
+  constructor(readonly props: Props, readonly errorHandler: (error: Error) => void) {
+    this.datasource = props.datasource;
+    this.onChange = props.onChange;
+    this.query = this.datasource.processQuery(props.query);
```
mure, line 20: I think because the DataFrameVariableQueryEditor uses this class, it will apply the query defaults to the variable model: https://github.com/ni/systemlink-grafana-plugins/blob/1dd516df02da54b709fef8d94131c478081f6bfd/src/datasources/data-frame/types.ts#L17
Which is a little weird because the `tableId` is the only property that matters for the variable query. It probably doesn't affect anything, but it is a little weird. Maybe we could move the `processQuery` call out of the common class.

mure (approved):

src/core/utils.ts
cameronwaterman: `current.value` isn't publicly accessible on the TypedVariableModel class. We do the same on line 51 of this file as well.

cameronwaterman:
@mure - Submitting this since teams is showing you as OOO and so Josh can start playing with it. If you have responses to those two comments I can address them in a separate PR.

PR #57, opened 2024-04-09
feat(tag): handle multiple tags

TigranVardanyan:
# Pull Request

## 🤨 Rationale

Need to show more than one tag values in single graph

## 👩‍💻 Implementation

Added functions to parse paths that contain multiple-value variables and return relevant data

## 🧪 Testing

tests have been added\changed for all changes in the code

## ✅ Checklist

@MelissaHilliard since instead of metadata/history of the first tag it now returns data from all matched tags, we need to change the tooltips

### DIFF
--- a/src/datasources/tag/TagDataSource.ts
+++ b/src/datasources/tag/TagDataSource.ts
@@ -1,17 +1,27 @@
 import {
   DataFrameDTO,
   DataSourceInstanceSettings,
-  dateTime,
   DataQueryRequest,
   TimeRange,
-  FieldDTO,
+  FieldConfig,
   FieldType,
+  MetricFindValue,
   TestDataSourceResponse,
 } from '@grafana/data';
 import { BackendSrv, TemplateSrv, getBackendSrv, getTemplateSrv } from '@grafana/runtime';
 import { DataSourceBase } from 'core/DataSourceBase';
-import { throwIfNullish } from 'core/utils';
-import { TagHistoryResponse, TagQuery, TagQueryType, TagsWithValues } from './types';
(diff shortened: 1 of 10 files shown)


src/datasources/tag/TagDataSource.ts
```diff
@@ -25,94 +35,222 @@ export class TagDataSource extends DataSourceBase<TagQuery> {
...
+  async metricFindQuery({ workspace, path }: TagVariableQuery): Promise<MetricFindValue[]> {
+    // Skip querying when no paths are provided
+    if (!path) {
+      return []
+    }
+    let paths: string[] = [path];
+    // wrap replaced variable in extra curly braces, protecting original {} tags from misinterpretation.
+    let parsed_workspace = this.templateSrv.replace(workspace,);
+    if (this.templateSrv.containsTemplate(path)) {
+      let replaced_path = this.templateSrv.replace(path, {}, (v: string) => `{{${v}}}`);
```
mure, line 208: Can you elaborate more on this and the comment above? Why do we need to add two sets of brackets?

src/datasources/tag/types.ts
```diff
@@ -47,11 +52,30 @@ export interface TagsWithValues {
   tagsWithValues: TagWithValue[];
 }
 
+export interface HttpHistoricalValue {
```
mure, line 55:
```suggestion
export interface TimestampedValue {
```

TigranVardanyan: usually I copy interface names from yml file
image

mure:
Closing in favor of #58

PR #58, opened 2024-04-24
feat(tag): visualizing current value of multiple tags

TigranVardanyan:
# Pull Request

## 🤨 Rationale

Need to visualize current value of multiple tags

## 👩‍💻 Implementation

Made little changes to handle multiple tags

## 🧪 Testing

Changed snapshots because structure of returned data is changed

## ✅ Checklist

### DIFF
--- a/src/datasources/tag/TagDataSource.test.ts
+++ b/src/datasources/tag/TagDataSource.test.ts
@@ -63,15 +63,37 @@ describe('queries', () => {
   });
 
   test('uses displayName property', async () => {
-    backendSrv.fetch.mockReturnValue(createQueryTagsResponse({ properties: { displayName: 'My cool tag' } }));
+    backendSrv.fetch.mockReturnValue(createQueryTagsResponse([{ tag: { properties: { displayName: 'My cool tag' } } }]));
 
     const result = await ds.query(buildQuery({ path: 'my.tag' }));
 
     expect(result.data).toMatchSnapshot();
   });
 
+  test('multiple current values with properties', async () => {
+    backendSrv.fetch.mockReturnValue(createQueryTagsResponse([
+      { tag: { path: 'my.1.tag', properties: { unit: 'A' } } },
+      { tag: { path: 'my.2.tag', properties: { unit: 'A' } }, current: { value: { value: '41.3' } } }
+    ]));
+
+    const result = await ds.query(buildQuery({ path: 'my.*.tag', properties: true }));
+
+    expect(result.data).toMatchSnapshot();
+  });
+
+  test('multiple current values with different properties', async () => {
+    backendSrv.fetch.mockReturnValue(createQueryTagsResponse([
+      { tag: { path: 'my.1.tag', properties: {  upperCriticalThreshold: '15'} } },
+      { tag: { path: 'my.2.tag', properties: { unit: 'A' } }, current: { value: { value: '41.3' } } }
+    ]));
+
+    const result = await ds.query(buildQuery({ path: 'my.*.tag', properties: true }));
+
+    expect(result.data).toMatchSnapshot();
+  });
+
   test('handles null tag properties', async () => {
-    backendSrv.fetch.mockReturnValue(createQueryTagsResponse({ properties: null }));
+    backendSrv.fetch.mockReturnValue(createQueryTagsResponse([{ tag: { properties: null } }]));
 
     const result = await ds.query(buildQuery({ path: 'my.tag' }));
 
@@ -79,7 +101,7 @@ describe('queries', () => {
   });
 
   test('handles tag with no current value', async () => {
-    backendSrv.fetch.mockReturnValue(createQueryTagsResponse({}, null));
+    backendSrv.fetch.mockReturnValue(createQueryTagsResponse([{ tag: {}, current: null }]));
 
     const result = await ds.query(buildQuery({ path: 'my.tag' }));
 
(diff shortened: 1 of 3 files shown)


src/datasources/tag/TagDataSource.ts
```diff
@@ -25,27 +25,53 @@ export class TagDataSource extends DataSourceBase<TagQuery> {
...
+    const tag = tagsLastUpdates[0].tag
     const name = tag.properties?.displayName ?? tag.path;
     const result: DataFrameDTO = { refId: query.refId, fields: [] };
 
     if (query.type === TagQueryType.Current) {
+      const allPossibleProps = this.getAllProperties(tagsLastUpdates);
       result.fields = [
-        { name, values: [this.convertTagValue(tag.type ?? tag.datatype, current?.value.value)] },
-        { name: 'updated', values: [current?.timestamp], type: FieldType.time, config: { unit: 'dateTimeFromNow' } },
-      ];
```
mure, line 48: Having separate fields for the name and value of a tag is technically a breaking change. The data frame structure changes as so:

| my.tag | updated      |
|--------|--------------|
| 3.14   | a minute ago |

-> 

| name   | currentValue | updated      |
|--------|--------------|--------------|
| my.tag | 3.14         | a minute ago |

While simple use cases should be unaffected, it's possible that this could break existing dashboards. We need to decide if we're okay with the breaking change or make the new behavior opt-in.

@joshuaprewitt @prestwick

src/datasources/tag/TagDataSource.ts
```diff
@@ -25,27 +25,53 @@ export class TagDataSource extends DataSourceBase<TagQuery> {
...
 
   async runQuery(query: TagQuery, { range, maxDataPoints, scopedVars }: DataQueryRequest): Promise<DataFrameDTO> {
-    const { tag, current } = await this.getLastUpdatedTag(
+    const tagsLastUpdates: TagWithValue[] = await this.getLastUpdatedTag(
       this.templateSrv.replace(query.path, scopedVars),
       this.templateSrv.replace(query.workspace, scopedVars)
     );
 
+    const tag = tagsLastUpdates[0].tag
     const name = tag.properties?.displayName ?? tag.path;
```
mure, line 42: Move this to the history branch of `runQuery` since it's no longer used in the `query.type === TagQueryType.Current` branch

src/datasources/tag/TagDataSource.ts
```diff
@@ -69,12 +95,12 @@ export class TagDataSource extends DataSourceBase<TagQuery> {
 
     const response = await this.post<TagsWithValues>(this.tagUrl + '/query-tags-with-values', {
       filter,
-      take: 1,
+      take: 10,
```
mure, line 98: I think 10 is a reasonable max number of tags to visualize at one time, but I would like PO feedback on this.

@prestwick @joshuaprewitt

src/datasources/tag/TagDataSource.ts
joshuaprewitt: I would recommend 16 or 32 so that a user could all of the CPUs for a single system or all of the analog inputs for a DAQ board.

src/datasources/tag/TagDataSource.ts
joshuaprewitt: I would recommend 16 or 32 so that a user could all of the CPUs for a single system or all of the analog inputs for a DAQ board.

src/datasources/tag/TagDataSource.ts
joshuaprewitt: If it is the right thing to do in the long run I’m ok with the breaking change.

joshuaprewitt: If we can get the change in early for the May release I would be happy to throughly test it out just to be safe.

src/datasources/tag/TagDataSource.ts
TigranVardanyan:
```suggestion
    if (query.properties) {
      result.fields = result.fields.concat(this.getPropertiesAsFields(tag.properties));
    }
```

But used here, outside of query.type check branch
Since we divided PR into 4 parts, I will delete this part in the second PR when I work with history case and will remove ability to show properties with history

src/datasources/tag/TagDataSource.ts
TigranVardanyan: As "/nitaghistorian/v2/tags/query-history" endpoint can return 10 tags at one query, difference between 16 and 32 will be 2 aditional queries, I think it is not critical. I will set it to 32

TJ-G: This is jogging my memory - I had thought the 10 item limitation was only on the frontend in SystemLink, but it looks like the team put additional protection in place on the backend as well. This limit was put in place to mitigate the current performance limitations of the tag service at scale. Hopefully in the future we can raise this limit...

How much extra work is it to add multiple queries here? I'm good starting with 10 in the first pass, but what's the expected delta in complexity?

src/datasources/tag/TagDataSource.ts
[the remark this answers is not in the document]
TigranVardanyan: I have not yet received any new descriptions from Vera Stoinova. Let's do it next time

src/datasources/tag/TagDataSource.ts
TigranVardanyan: @TJ-G  function to fetch historical data in chunks of 10 tags is ready

src/datasources/tag/TagDataSource.test.ts
```diff
@@ -63,23 +63,45 @@ describe('queries', () => {
...
+    ]));
+
+    const result = await ds.query(buildQuery({ path: 'my.*.tag', properties: true }));
+
+    expect(result.data).toMatchSnapshot();
+  });
+
+  test('multiple current values with different properties', async () => {
+    backendSrv.fetch.mockReturnValue(createQueryTagsResponse([
+      { tag: { path: 'my.1.tag', properties: {  upperCriticalThreshold: '15'} } },
```
mure, line 86:
```suggestion
      { tag: { path: 'my.1.tag', properties: { upperCriticalThreshold: '15' } } },
```

nit: spacing

mure (approved):

PR #60, opened 2024-05-15
feat(tag): visualize history of multiple tags

TigranVardanyan:
# Pull Request

## 🤨 Rationale

The "history" query type currently visualizes data only for the first matched tag. Need to visualize data from all matched tags.

## 👩‍💻 Implementation

Implemented functions to retrieve tag history in chunks and merge responses into a single dataframe. Since it is impossible to have properties of multiple tags in one dataframe with those historical values, "Properties" option in now available only for query type "Current".

## 🧪 Testing

Added a test to check if the history of multiple tags is processed correctly. Added a test to ensure the properties switch button is hidden when the query type is "History".

## ✅ Checklist

### DIFF
--- a/src/datasources/tag/TagDataSource.ts
+++ b/src/datasources/tag/TagDataSource.ts
@@ -81,15 +81,25 @@ export class TagDataSource extends DataSourceBase<TagQuery> {
       const workspaceTagMap: Record<string, TagWithValue[]> = {};
       const tagPropertiesMap: Record<string, Record<string, string> | null> = {};
 
-      tagsWithValues.forEach(tagWithValue => {
-        let workspace = tagWithValue.tag.workspace ?? tagWithValue.tag.workspace_id
+      // Identify tags that exist in more than one workspace
+      const tagPathCount: Record<string, number> = {};
+      for (const tagWithValue of tagsWithValues) {
+        tagPathCount[tagWithValue.tag.path] = (tagPathCount[tagWithValue.tag.path] || 0) + 1;
+      }
+
+      for (const tagWithValue of tagsWithValues) {
+        const workspace = tagWithValue.tag.workspace ?? tagWithValue.tag.workspace_id;
         if (!workspaceTagMap[workspace]) {
           workspaceTagMap[workspace] = [];
         }
         workspaceTagMap[workspace].push(tagWithValue);
-        tagPropertiesMap[`${getWorkspaceName(workspaces, workspace)}.${tagWithValue.tag.path}`] = tagWithValue.tag.properties
-      });
-      let tagsDecimatedHistory: { [key: string]: TypeAndValues } = {}
+        const prefixedPath = tagPathCount[tagWithValue.tag.path] > 1
+            ? `${getWorkspaceName(workspaces, workspace)}.${tagWithValue.tag.path}`
+            : tagWithValue.tag.path;
+        tagPropertiesMap[prefixedPath] = tagWithValue.tag.properties;
+      };
+
+      let tagsDecimatedHistory: { [key: string]: TypeAndValues } = {};
       for (const workspace in workspaceTagMap) {
         const tagHistoryResponse = await this.getTagHistoryWithChunks(
           workspaceTagMap[workspace],
@@ -98,18 +108,16 @@ export class TagDataSource extends DataSourceBase<TagQuery> {
           maxDataPoints
         )
         for (const path in tagHistoryResponse.results) {
-          // If tags are mixed from different workspaces, add the workspace name as a prefix
-          if (Object.keys(workspaceTagMap).length === 1) {
-            tagsDecimatedHistory[path] = tagHistoryResponse.results[path]
-          } else {
-            tagsDecimatedHistory[`${getWorkspaceName(await this.getWorkspaces(), workspace)}.${path}`] = tagHistoryResponse.results[path]
-          }
+          const prefixedPath = tagPathCount[path] > 1
+            ? `${getWorkspaceName(workspaces, workspace)}.${path}`
+            : path;
+          tagsDecimatedHistory[prefixedPath] = tagHistoryResponse.results[path];
         }
       }
 
       const mergedTagValuesWithType = this.mergeTagsHistoryValues(tagsDecimatedHistory);
       result.fields.push({
-        name: 'time', 'values': mergedTagValuesWithType.timestamps, type: FieldType.time
+        name: 'time', values: mergedTagValuesWithType.timestamps, type: FieldType.time
       });
 
       for (const path in mergedTagValuesWithType.values) {
--- a/src/datasources/tag/TagDataSource.ts
+++ b/src/datasources/tag/TagDataSource.ts
@@ -6,6 +6,7 @@ import {
   FieldType,
   TestDataSourceResponse,
   FieldConfig,
+  dateTime,
 } from '@grafana/data';
 import { BackendSrv, TemplateSrv, getBackendSrv, getTemplateSrv } from '@grafana/runtime';
 import { DataSourceBase } from 'core/DataSourceBase';
@@ -94,10 +95,10 @@ export class TagDataSource extends DataSourceBase<TagQuery> {
         }
         workspaceTagMap[workspace].push(tagWithValue);
         const prefixedPath = tagPathCount[tagWithValue.tag.path] > 1
-            ? `${getWorkspaceName(workspaces, workspace)}.${tagWithValue.tag.path}`
-            : tagWithValue.tag.path;
+          ? `${getWorkspaceName(workspaces, workspace)}.${tagWithValue.tag.path}`
+          : tagWithValue.tag.path;
         tagPropertiesMap[prefixedPath] = tagWithValue.tag.properties;
-      };
+      }
 
       let tagsDecimatedHistory: { [key: string]: TypeAndValues } = {};
       for (const workspace in workspaceTagMap) {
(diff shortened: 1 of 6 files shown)


src/datasources/tag/components/TagQueryEditor.tsx
```diff
@@ -51,19 +50,21 @@ export function TagQueryEditor({ query, onChange, onRunQuery, datasource }: Prop
...
 
 const tooltips = {
-  queryType: `Current allows you to visualize the most recent tag value. History allows you to
-              visualize tag values over time. Historical values use the time range set on the
+  queryType: `Current allows you to visualize the most recent value of multiple tags. History allows you to
+              visualize the values of multiple tags over time. Historical values use the time range set on the
               dashboard and are decimated according to the width of the panel.`,
 
-  tagPath: `The full path of the tag to visualize. You can enter a variable into this field.`,
+  tagPath: `The path to search for the tags you want to visualize. You can enter a variable into this field and use glob-style patterns as wildcards.`,
```
mure, line 67: @vpstoynova Could you review these strings? My personal opinion:

"Current allows you to visualize the most recent value of **one or more tags**"

"The path to search for the tags you want to visualize. **Use * as a wildcard to match multiple tags.**"

src/datasources/tag/TagDataSource.ts
```diff
@@ -73,18 +78,43 @@ export class TagDataSource extends DataSourceBase<TagQuery> {
...
-      ];
-    }
+      const tagPropertiesMap: Record<string, Record<string, string> | null> = {};
+      tagsWithValues.forEach((tag: TagWithValue) => {
+        tagPropertiesMap[tag.tag.path] = tag.tag.properties
+      });
+      const workspaceFromFirstTag = tagsWithValues[0].tag.workspace ?? tagsWithValues[0].tag.workspace_id;
+      const tagHistoryResponse = await this.getTagHistoryWithChunks(
+        Object.keys(tagPropertiesMap),
+        workspaceFromFirstTag,
```
mure, line 88: So when visualizing the current value of tags, the matched tags can be from a mixed set of workspaces, but for historical values, it's locked to the workspace of the first matched tag. I think this could be confusing for users. We may need to force users to explicitly pick a workspace or make separate requests for each unique workspace.

@TJ-G

TigranVardanyan: I can group tags by workspace, change **getTagHistoryWithChunks** signature and as you said, make separate requests for each workspace.

So, in case we have tags with the same path (or displayName), should I prefix the workspace name to the display names ​​so that they can be differentiated?

@mure @TJ-G @joshuaprewitt

TJ-G: Where is this displayed to the user? Are we talking about in the panel legend?
image

If so, yeah a format like {workspace}.{tagpath} could be a good option.

src/datasources/tag/components/TagQueryEditor.tsx
vpstoynova: I think your suggestions look good.

src/datasources/tag/TagDataSource.ts
markacamps: @TigranVardanyan are there any additional PRs for AAU open?  If not and this is the only one, are we complete once this is merged?

TigranVardanyan: @markacamps there are 2 more features that I would like to add, but they are not required for AAU

src/datasources/tag/TagDataSource.ts
```diff
@@ -73,18 +78,61 @@ export class TagDataSource extends DataSourceBase<TagQuery> {
...
+          workspace,
+          range,
+          maxDataPoints
+        )
+        for (const path in tagHistoryResponse.results) {
+          // If tags are mixed from different workspaces, add the workspace name as a prefix
+          if (Object.keys(workspaceTagMap).length === 1) {
+            tagsDecimatedHistory[path] = tagHistoryResponse.results[path]
+          } else {
+            tagsDecimatedHistory[`${getWorkspaceName(await this.getWorkspaces(), workspace)}.${path}`] = tagHistoryResponse.results[path]
```
mure, line 105: I think we should only format it as `workspace.path` if there are duplicate paths. We should also consider doing this for the "Current" query to be consistent.

markacamps: @TigranVardanyan can you make this change?

TigranVardanyan: @markacamps done
@mure for current we can add additional column for workspace and when it will be necessary to combine with the name through the Grafana transformations

src/datasources/tag/components/TagQueryEditor.tsx
mure: @TigranVardanyan it doesn't look like this feedback was ever incorporated

TigranVardanyan: @mure updated

mure (changes requested):
Screenshot 2024-08-30 170232

I tested this out locally and the data isn't looking like how I would expect. Even when the timestamps match up, the columns alternate missing values, which seems to create gaps in the chart.

Screenshot 2024-08-30 173523

It gets weirder when your tags aren't updating at the same frequency.
There's not any way to guarantee that the tag history returned by the query will have uniform timestamps, so I think what we should do is return each tag in a separate data frame/series (the second example in [these docs](https://grafana.com/developers/plugin-tools/key-concepts/data-frames#wide-format)).

This is the first time we've needed to do this in a plugin, so we'll need to make the `runQuery` method support returning a list of data frames: `DataFrameDTO[]`

One advantage of this approach is that we can continue to have the **Properties** toggle for the history query instead of hiding it.

src/datasources/tag/__snapshots__/TagDataSource.test.ts.snap
```diff
@@ -430,12 +430,14 @@ exports[`queries numeric tag history 1`] = `
     "fields": [
       {
         "name": "time",
+        "type": "time",
         "values": [
-          1672531200000,
-          1672531260000,
```
mure, line 435: This shouldn't be changing. In my experience, some Grafana visualizations don't handle string timestamps correctly. We should be converting the timestamps to unix milliseconds.

mure (changes requested):

src/datasources/tag/__snapshots__/TagDataSource.test.ts.snap
markacamps: @mure We reviewed your suggestions and understand your perspective, but we also have this working in SLS as Tigran currently has the runQuery returning what the visualization requires.  Given our time constraints, is this necessary to move forward or can we move forward as-is?

mure: @mure If we've tested this in SLS and it satisfies the requirements for AAU, then I won't block the review on this.

mure (approved):

TJ-G (approved):

cameronwaterman (changes requested):
Moving to "Request changes" in order to review open conversations

src/datasources/tag/TagDataSource.ts
cameronwaterman: @markacamps @TigranVardanyan - This conversation is unresolved and looks like it might require a response to TJ's question.

src/datasources/tag/TagDataSource.ts
cameronwaterman: @TigranVardanyan - If the change has been made, can this conversation be resolved?

TigranVardanyan: yes

src/datasources/tag/__snapshots__/TagDataSource.test.ts.snap
cameronwaterman: @TigranVardanyan @markacamps 

If the correct thing to do is convert these timestamps to unix milliseconds, that is what we should do. It shouldn't be a large amount of work, either. We need to remember that SLS isn't the only client of these plugins and just because it works there doesn't mean it does everywhere or in every edge-case scenario.

cameronwaterman:
@markacamps @TigranVardanyan - Carson left a comment yesterday related to the return type of the `runQuery` function. Please take a look as it doesn't appear that comment has been resolved.

src/datasources/tag/TagDataSource.ts
TigranVardanyan: not required for AAU, we can close this one

src/datasources/tag/__snapshots__/TagDataSource.test.ts.snap
TigranVardanyan: @cameronwaterman I will fix milliseconds part. What about gaps in the chart, it fixes by **Connect null values** option
Earlier I tried many times to return separated data frame for each timeframe, but with no success, only way I founded to mix frames to one and fill it with null values

mure: I still think multiple data frames is the better approach, but with that workaround I'm okay with submitting this for now

mure:
@cameronwaterman 
> Carson left a comment yesterday related to the return type of the runQuery function. Please take a look as it doesn't appear that comment has been resolved

We won't need to do this if we're only returning one data frame, as was the outcome of the other conversation

src/datasources/tag/__snapshots__/TagDataSource.test.ts.snap
TigranVardanyan: @cameronwaterman  @mure done

markacamps (approved):

cameronwaterman (approved):

markacamps:
@mure is this clear to merge?

mure:
> @mure is this clear to merge?

Done

PR #62, opened 2024-06-12
feat(asset): add asset data source

TigranVardanyan:
move error component to core
test for error component
move getSystems to base datasource
minor changes

# Pull Request

## 🤨 Rationale

Asset data source

## 👩‍💻 Implementation

Asset datasource use Asset Performance Management API to retrieve and show asset-related data

## 🧪 Testing

Includes unit tests to ensure stable usage of the data source.

## ✅ Checklist


mure (changes requested):
If a system doesn't have an alias, we need to display its ID as a fallback:

image

src/datasources/asset/types.ts
```diff
@@ -0,0 +1,149 @@
...
+  supportsReset: boolean,
+  id: string,
+  location: AssetLocationModel,
+  calibrationStatus: 'OK' | 'APPROACHING_RECOMMENDED_DUE_DATE' | 'PAST_RECOMMENDED_DUE_DATE',
+  isSystemController: boolean,
+  externalCalibration?: ExternalCalibrationModel,
+  discoveryType: 'MANUAL' | 'AUTOMATIC',
+  properties: {
+    [key: string]: string
+  },
```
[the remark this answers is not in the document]
TigranVardanyan: changed everywhere

src/datasources/asset/components/AssetQueryEditor.tsx
```diff
@@ -0,0 +1,156 @@
...
+    }
+  };
+
+  const onWorkspaceChange = (item?: SelectableValue<string>): void => {
+    if (item?.value) {
+      // if workspace changed, reset Systems and Assets fields
+      handleQueryChange(
+        { ...query, workspace: item.value, minionIds: [] },
+        // do not run query if workspace not changed
+        query.workspace !== item.value
```
mure, line 48: I think it would make more sense to push this up to the if statement condition, because we don't need to call `onChange` if the workspace is the same.

`if (item?.value && item.value !== query.workspace)`

src/datasources/asset/components/AssetQueryEditor.tsx
```diff
@@ -0,0 +1,156 @@
...
+      handleQueryChange({ ...query, workspace: '' }, true);
+    }
+  };
+  const handleMinionIdChange = (items: Array<SelectableValue<string>>): void => {
+    if (items) {
+      handleQueryChange(
+        { ...query, minionIds: items.map(i => i.value!) },
+        // do not run query if minionIds not changed
+        !_.isEqual(query.minionIds, items)
+      );
```
mure, line 60: My other comment applies here as well

src/datasources/asset/components/AssetQueryEditor.tsx
```diff
@@ -0,0 +1,156 @@
...
+          onChange={onWorkspaceChange}
+          options={workspaces.value}
+          placeholder="Any workspace"
+          value={query.workspace}
+        />
+      </InlineField>
+      <InlineField label="Systems"
+                   tooltip={tooltips.system[EntityType.Asset]}
+                   labelWidth={22}
+                   data-testid={selectors.components.assetPlugin.system}>
```
mure, line 101: Why do we need the testid attribute here? Can we use `getByLabelText` like we do here: https://github.com/ni/systemlink-grafana-plugins/blob/0e230a372d46e42326bde813bccf74e24c4ed0a8/src/datasources/tag/components/TagQueryEditor.test.tsx#L17

src/datasources/asset/components/AssetQueryEditor.test.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+const workspacesLoaded = () => waitForElementToBeRemoved(screen.getByTestId('Spinner'));
+
+test('renders with query defaults', async () => {
+  render({} as AssetQuery)
+  await workspacesLoaded()
+
+  expect(screen.getByTestId(selectors.components.assetPlugin.workspace)).toHaveTextContent('Any workspace');
+  expect(screen.getByTestId(selectors.components.assetPlugin.system)).toHaveTextContent('Select systems');
+})
```
mure, line 18: We should add a test like [this one](https://github.com/ni/systemlink-grafana-plugins/blob/0e230a372d46e42326bde813bccf74e24c4ed0a8/src/datasources/tag/components/TagQueryEditor.test.tsx#L22) that verifies that the UI renders with the user's saved query.

src/datasources/asset/components/AssetQueryEditor.tsx
TigranVardanyan: getByLabelText not work with Select component
image

_TestingLibraryElementError: Found a label with the text of: Workspace, however the element associated with this label (div) is non-labellable [https://html.spec.whatwg.org/multipage/forms.html#category-label]. If you really need to label a div, you can use aria-label or aria-labelledby instead._

getByRole

src/datasources/asset/AssetDataSource.ts
```diff
@@ -0,0 +1,111 @@
...
+  async runQuery(query: AssetQuery, options: DataQueryRequest): Promise<DataFrameDTO> {
+    return await this.processMetadataQuery(query)
+  }
+
+  async processMetadataQuery(query: AssetQuery) {
+    const result: DataFrameDTO = { refId: query.refId, fields: [] };
+    let workspaceId = this.templateSrv.replace(query.workspace);
+    const minionIds = replaceVariables(query.minionIds, this.templateSrv);
+    const conditions = [];
+    if (workspaceId) {
```
CiprianAnton, line 45: Minor: you could move workspaceId query after the minionIds, since workspace is not indexed

src/datasources/asset/AssetDataSource.ts
```diff
@@ -0,0 +1,111 @@
...
+      { name: 'bus type', values: assets.map(a => a.busType) },
+      { name: 'asset type', values: assets.map(a => a.assetType) },
+      { name: 'is NI asset', values: assets.map(a => a.isNIAsset) },
+      { name: 'calibration status', values: assets.map(a => a.calibrationStatus) },
+      { name: 'is system controller', values: assets.map(a => a.isSystemController) },
+      { name: 'last updated timestamp', values: assets.map(a => a.lastUpdatedTimestamp) },
+      { name: 'minionId', values: assets.map(a => a.location.minionId) },
+      { name: 'parent name', values: assets.map(a => a.location.parent) },
+      { name: 'system name', values: assets.map(a => a.location.systemName) },
+      {
```
CiprianAnton, line 69: Is this something necessary now? It only uses NextRecommendedDate and does not take into account the NextCustomDueDate (overwrite for the first one). We plan to add better support for this

src/datasources/asset/types.ts
```diff
@@ -0,0 +1,140 @@
...
+  PRESENT = "PRESENT"
+}
+
+export enum ExternalCalibrationEntryType {
+  AUTOMATIC = 'AUTOMATIC',
+  MANUAL = 'MANUAL'
+}
+
+export enum SystemConnection {
+  APPROVED = "APPROVED",
```
CiprianAnton, line 132: In SLE, system connection returned by asset has only 2 values:
DISCONNECTED
CONNECTED

src/datasources/asset/types.ts
```diff
@@ -0,0 +1,140 @@
...
+  totalCount: number
+}
+
+export interface AssetModel {
+  modelName: string,
+  modelNumber: number,
+  serialNumber: string,
+  vendorName: string,
+  vendorNumber: number,
+  busType: 'BUILT_IN_SYSTEM' | 'PCI_PXI' | 'USB' | 'GPIB' | 'VXI' | 'SERIAL' | 'TCP_IP' | 'CRIO' | 'SCXI' | 'CDAQ' | 'SWITCH_BLOCK' | 'SCC' | 'FIRE_WIRE' | 'ACCESSORY' | 'CAN' | 'SWITCH_BLOCK_DEVICE' | 'SLSC',
```
CiprianAnton, line 29: What happens if we add a new bus type in the service? I guess it involves updating the plugin. My question is whether we define this as a generic string, so the client is more flexible for modifications. I think this is possible if the code is not doing particular checks with these values.

This applies for other enums as well.

src/datasources/asset/types.ts
```diff
@@ -0,0 +1,140 @@
...
+  assetPresence: AssetPresence,
+  systemConnection?: SystemConnection
+}
+
+export interface AssetLocationModel {
+  minionId: string,
+  parent: string,
+  resourceUri: string,
+  slotNumber: number,
+  systemName: string,
```
CiprianAnton, line 66: Not returned by asset anymore. What is missing from here is physicalLocation that gets populated when the asset is in an unmanaged location (e.g. cabinet).

src/datasources/asset/types.ts
```diff
@@ -0,0 +1,140 @@
+import { DataQuery } from '@grafana/schema'
```
CiprianAnton, line 1: General question about this api, is everything defined here being used? For e.g., the only filters I've seen in the code is around location and workspace. If that is the case, I think this file can be simplified

src/datasources/asset/AssetDataSource.ts
TigranVardanyan: ok, I'll delete this line

src/datasources/asset/types.ts
TigranVardanyan: will add interface for SLE and merge them like TagWithValue interface for SLS, SLE and SystemLink Cloud

src/datasources/asset/types.ts
TigranVardanyan: can we just add string type in the end?

src/datasources/asset/types.ts
TigranVardanyan: @CiprianAnton did not catch, what exact is not returned anymore?

src/datasources/asset/types.ts
TigranVardanyan: everything will be used for features that already done but not exists in current PR

src/datasources/asset/types.ts
CiprianAnton: Yes, I agree

src/datasources/asset/types.ts
CiprianAnton: systemName

src/datasources/asset/AssetDataSource.ts
```diff
@@ -0,0 +1,111 @@
...
+      })
+
+      return response.data;
+    } catch (error) {
+      throw new Error(`An error occurred while querying systems: ${error}`);
+    }
+  }
+
+  async testDatasource(): Promise<TestDataSourceResponse> {
+    await this.get(this.baseUrl + '/assets?take=1');
```
kkerezsi, line 108: Is this GET going to throw an error in case of 4xx / 5xx status codes ?

Shouldn't we at least check that the response is a valid result before returning success ?

src/core/errors.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { act } from 'react-dom/test-utils';
+import { FloatingError, parseErrorMessage } from './errors';
+import { SystemLinkError } from "./types";
+import React from 'react';
+import { errorCodes } from "../datasources/data-frame/constants";
+
+test('renders with error message', () => {
+  const { container } = render(<FloatingError message='error msg'/>)
+
+  expect(container.innerHTML).toBeTruthy() // refact: get by text
```
kkerezsi, line 12: I see some comments here for refactoring . Is this something you are going to do in this PR or another one ?

src/datasources/asset/__snapshots__/AssetDataSource.test.ts.snap
```diff
@@ -0,0 +1,146 @@
...
+        "name": "system name",
+        "values": [
+          "NI-sbRIO-9629-01FE20D1",
+          "NI-sbRIO-9629-01FE20D1",
+          "NI-sbRIO-9629-01FE20D1",
+          "NI-sbRIO-9629-01FE20D1",
+        ],
+      },
+      {
+        "name": "calibration due date",
```
kkerezsi, line 125: Is this something we want to keep here ? It has 4 null values.

src/datasources/asset/AssetDataSource.ts
```diff
@@ -0,0 +1,111 @@
...
+  AssetFilterProperties,
+  AssetModel, AssetQuery,
+  AssetQueryType,
+  AssetsResponse,
+} from './types';
+import { getWorkspaceName, replaceVariables } from "../../core/utils";
+import { SystemMetadata } from "../system/types";
+import { defaultOrderBy, defaultProjection } from "../system/constants";
+
+export class AssetDataSource extends DataSourceBase<AssetQuery> {
```
kkerezsi, line 19: This is more of a question .. but I am wondering , why aren't we using custom hooks for the datasources if we are using react components ? Are there some DI related reasons ?

kkerezsi (commented):
I've added couple of small comments.

src/datasources/asset/__snapshots__/AssetDataSource.test.ts.snap
TigranVardanyan: taking into account this comment (https://github.com/ni/systemlink-grafana-plugins/pull/62#discussion_r1654551205), I will remove that part

src/datasources/asset/AssetDataSource.ts
```diff
@@ -0,0 +1,111 @@
...
+        orderBy: defaultOrderBy,
+      })
+
+      return response.data;
+    } catch (error) {
+      throw new Error(`An error occurred while querying systems: ${error}`);
+    }
+  }
+
+  async testDatasource(): Promise<TestDataSourceResponse> {
```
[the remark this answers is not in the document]
TigranVardanyan: image
this one is abstract function from DataSourceApi class

src/datasources/asset/AssetDataSource.ts
TigranVardanyan: I'm sure the testDatasource function is only runs when we check the validity of the credentials, it throws a 401 when the user is not authorized, same as rest of testDatasource in our other plugins

src/datasources/asset/AssetDataSource.ts
TigranVardanyan: I follow the way the other plugins are written

src/datasources/asset/components/AssetQueryEditor.tsx
TigranVardanyan: best solution I founded is find all combo boxes with getAllByRole() and take them by index

mure: I think that's fine for now

src/core/errors.test.tsx
TigranVardanyan: done

src/datasources/asset/types.ts
TigranVardanyan: removed, added physicalLocation as string to AssetLocationModel interface

src/datasources/asset/types.ts
TigranVardanyan: all enums that will not be used in code(if-else statments or etc) are replaced with strings

src/datasources/asset/types.ts
TigranVardanyan: see below comment

src/datasources/asset/components/AssetQueryEditor.test.tsx
TigranVardanyan: added test _renders with initial query and updates when user makes changes_

src/datasources/asset/AssetDataSource.ts
```diff
@@ -0,0 +1,106 @@
...
+    ];
+    return result;
+  }
+
+
+  shouldRunQuery(_: AssetQuery): boolean {
+    return true;
+  }
+
+  async queryAssets(filter = '', take = -1): Promise<AssetModel[]> {
```
CiprianAnton, line 78: take=-1 might not work when there are many assets (e.g. 100k).
One possible solution is to get data in batches:
- Request1: skip 0, take 5000
- Request2: skip 5000, take 5000

I'm not saying we need to handle this now but might worth having a work item to address this in the future.

src/datasources/asset/types.ts
CiprianAnton: So this is something that needs work with both SLS and SLE, reason you keep old enum values as well?

TigranVardanyan: retained both values ​​to support SLS and SLE, added comment

mure (approved):

kkerezsi (approved):
Please check with @CiprianAnton for further feedback on assets as needed.

PR #68, opened 2024-09-17
feat(asset): add calibration forecast property only row formatting

kkerezsi:
# Pull Request

## 🤨 Rationale

When rendering property values , we want to format the data structure so that timecharts can also render property groupings.

[Task 2842497](https://dev.azure.com/ni/DevCentral/_workitems/edit/2842497): Implement Row view when non-time values only are selected

## 👩‍💻 Implementation

- Implemented method to format DataFrame structure to row view
- Ensure that time column is up to date with the latest values

## 🧪 Testing

- Unit testing
- Manual testing

## ✅ Checklist

### DIFF
--- a/src/datasources/asset-calibration/AssetCalibrationDataSource.test.ts
+++ b/src/datasources/asset-calibration/AssetCalibrationDataSource.test.ts
@@ -72,6 +72,14 @@ const modelGroupCalibrationForecastResponseMock: CalibrationForecastResponse =
     }
 }
 
+const emptyGroupCalibrationForecastResponseMock: CalibrationForecastResponse =
+{
+    calibrationForecast: {
+        columns: [
+        ]
+    }
+}
+
 const modelLocationGroupCalibrationForecastResponseMock: CalibrationForecastResponse =
 {
     calibrationForecast: {
@@ -221,6 +229,16 @@ describe('queries', () => {
       .mockReturnValue(createFetchResponse(monthLocationGroupCalibrationForecastResponseMock as CalibrationForecastResponse))
 
     const result = await datastore.query(buildCalibrationForecastQuery(monthLocationBasedCalibrationForecastQueryMock))
+
+    expect(result.data).toMatchSnapshot()
+  })
+
+  test('calibration forecast with month groupBy returns empty results', async () => {
+    backendServer.fetch
+      .calledWith(requestMatching({ url: '/niapm/v1/assets/calibration-forecast' }))
+      .mockReturnValue(createFetchResponse(emptyGroupCalibrationForecastResponseMock as CalibrationForecastResponse))
+
+    const result = await datastore.query(buildCalibrationForecastQuery(monthBasedCalibrationForecastQueryMock))
 
     expect(result.data).toMatchSnapshot()
   })
--- a/src/datasources/asset-calibration/AssetCalibrationDataSource.test.ts
+++ b/src/datasources/asset-calibration/AssetCalibrationDataSource.test.ts
@@ -9,15 +9,16 @@ import {
 } from "test/fixtures";
 import { AssetCalibrationDataSource } from "./AssetCalibrationDataSource";
 import {
-  AssetCalibrationForecastGroupByType,
+  AssetCalibrationPropertyGroupByType,
   AssetCalibrationQuery,
+  AssetCalibrationTimeBasedGroupByType,
   CalibrationForecastResponse,
 } from "./types";
 
-let ds: AssetCalibrationDataSource, backendSrv: MockProxy<BackendSrv>
+let datastore: AssetCalibrationDataSource, backendServer: MockProxy<BackendSrv>
 
 beforeEach(() => {
-  [ds, backendSrv] = setupDataSource(AssetCalibrationDataSource);
+  [datastore, backendServer] = setupDataSource(AssetCalibrationDataSource);
 });
 
 const monthGroupCalibrationForecastResponseMock: CalibrationForecastResponse =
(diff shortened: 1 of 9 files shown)


src/datasources/asset-calibration/AssetCalibrationDataSource.ts
```diff
@@ -45,18 +45,40 @@ export class AssetCalibrationDataSource extends DataSourceBase<AssetCalibrationQ
...
+
+      return result.fields;
+    }
+
+    const formattedFields = [] as FieldDTO[];
+    formattedFields.push({ name: "Group", values: [] } as FieldDTO);
+    formattedFields.push({ name: "Assets", values: [] } as FieldDTO);
+
+    for (let columnIndex = 0; columnIndex < result.fields.length; columnIndex++) {
+      formattedFields[0].values?.push(result.fields[columnIndex].name)
```
CiprianAnton, line 66: Values is non nullable, I suggest removing `?`

CiprianAnton (approved):

src/datasources/asset-calibration/AssetCalibrationDataSource.test.ts
```diff
@@ -0,0 +1,107 @@
...
+const locationGroupCalibrationForecastResponseMock: CalibrationForecastResponse =
+{
+    calibrationForecast: {
+        columns: [
+          { name: "Location1", values: [1] },
+          { name: "Location2", values: [2] },
+          { name: "Location3", values: [3] },
+        ]
+    }
+}
```
mure, line 42: It's kind of weird that these response formats are different orientations. Is this API still in development? Why isn't the second response:

```js
calibrationForecast: {
    columns: [
      { name: "Location", values: ["Location 1", "Location 2", "Location 3"] },
      { name: "Assets", values: [1, 2, 3] }
    ]
}
```

src/datasources/asset-calibration/AssetCalibrationDataSource.ts
```diff
@@ -45,18 +45,40 @@ export class AssetCalibrationDataSource extends DataSourceBase<AssetCalibrationQ
 
     result.fields = calibrationForecastResponse.calibrationForecast.columns || [];
     result.fields = result.fields.map(field => this.formatField(field, query));
+    result.fields = this.formatResultsWhenNoTimeSpecified(result, query);
```
mure, line 48: This logic is kind of odd to me. Line 47 does nothing if it's a location/model grouping, and line 48 does nothing if it's a time-based grouping. I think there should be two different "pipelines" depending on which response format we get, something like:

```js
if (isGroupedByTime(query)) {
  this.processResultsGroupedByTime(...)
} else {
  this.processResultsGroupedByCategory(...)
}

File: src/datasources/asset-calibration/AssetCalibrationDataSource.ts [hunk 4]
```diff
@@ -74,6 +96,8 @@ export class AssetCalibrationDataSource extends DataSourceBase<AssetCalibrationQ
       [AssetCalibrationForecastGroupByType.Day]: this.formatDateForDay,
       [AssetCalibrationForecastGroupByType.Week]: this.formatDateForWeek,
       [AssetCalibrationForecastGroupByType.Month]: this.formatDateForMonth,
+      [AssetCalibrationForecastGroupByType.Location]: (v: string) => v,
+      [AssetCalibrationForecastGroupByType.Model]: (v: string) => v,
```
mure, line 100: It doesn't look like these will ever be called because line 103 has `formatFunctionMap[timeGrouping!]`

mure (commented):
Sorry for missing the first review for the calibration forecast stuff. 

Some notes:
We're not configuring the day/week/month fields to be [FieldType.Time](https://github.com/ni/systemlink-grafana-plugins/blob/3a31c5b7346796eadcc6843149ba423b33955d60/src/datasources/tag/TagDataSource.ts#L65), so it doesn't work with the time series visualization:

image

Is this intended?

I also noticed that the Week and Month formatting isn't happening at all:

image

src/datasources/asset-calibration/AssetCalibrationDataSource.test.ts
kkerezsi: @mure we discussed if this should be an API level change or it should be a plugin implementation.

The way we are returning the data is the following:

calibrationForecast: {
    columns: [
      { name: "Location1", values: [1] },
      { name: "Location2", values: [2] }
      ....
    ]
}

To enable rendering the data on a bar chart, we want to ensure that the structure has the right format. Since this is a client specific need, we can implement it as part of the client code.

If you have concerns, please let me know and we can take it on a call

kkerezsi: @mure one more piece of detail that I have mentioned in the HLD:

The forecast can return 3 types of aggregations:

1. Time based 
    calibrationForecast: {
        columns: [
            { name: "Month", values: ["2022-01-01T00:00:00.0000000Z", "2022-02-01T00:00:00.0000000Z", "2022-03-01T00:00:00.0000000Z"] },
            { name: "Assets", values: [1, 0, 3] },
        ]
    }

2. Property based:

calibrationForecast: {
columns: [
{ name: "Location1", values: [1] },
{ name: "Location2", values: [2] }
....
]
}

3. Time & property based

calibrationForecast: {
columns: [
 { name: "Month", values: ["2022-01-01T00:00:00.0000000Z", "2022-02-01T00:00:00.0000000Z", "2022-03-01T00:00:00.0000000Z"] },
{ name: "Location1", values: [1, 2, 3] },
{ name: "Location2", values: [2, 3, 1] }
....
]
}


As you can see, the columns have a certain rule on how they are formatted. Having us provide a specific format from the backend only to satisfy a rendering option that the ui needs isn't something we want to do.

src/datasources/asset-calibration/AssetCalibrationDataSource.ts
kkerezsi: @mure I agree with you, we can write it in a more intuitive way.

src/datasources/asset-calibration/AssetCalibrationDataSource.ts
kkerezsi: Will remove these

mure (approved):

PR #69, opened 2024-09-24
feat(asset): Added calibration forecast query builder

saradei-ni:
# Pull Request

## 🤨 Rationale

[User Story 2798210](https://ni.visualstudio.com/DevCentral/_workitems/edit/2798210): [UI][Grafana] Add support for filters on the Grafana plugin

## 👩‍💻 Implementation

- integrated the query builder component
- moved constants and types that are used in multiple places in core
- created and integrated callbacks for query builder expression templates in order to handle computed field queries (ex: Location.MinionId || Location.PhysicalLocation)
- updated expression template callbacks to take option label/value into consideration so that the label is displayed and the value is sent

## 🧪 Testing

Added unit tests.

GIF  of the query builder sending the right requests.
test

## ✅ Checklist

- [x ] This PR has a title that follows the [commit message format](https://github.com/ni/systemlink-grafana-plugins#commit-message-format).


src/core/query-builder.utils.ts
```diff
@@ -0,0 +1,76 @@
...
+        fieldName = computedFieldName;
+        break;
+      }
+    }
+
+    const fieldOptions = options[fieldName];
+
+    if (fieldOptions?.length) {
+      const valueLabel = fieldOptions.find(option => option.value === value);
+
```
kkerezsi, line 67: nit: I like when the code gets a bit of space to breathe , however too many empty lines can also cause slight difficulty in reading.

My recommendation is to use empty lines only between different logical blocks. For instance the declaration of fieldOptions and the if check are part of the same logical block.

src/core/query-builder.utils.test.ts
```diff
@@ -0,0 +1,111 @@
...
+    Obj1: 'obj1.prop1 || obj1.prop2',
+    Obj2: 'obj2.prop1 || obj2.extra.prop2 || obj2.prop3 || obj2.prop4 || obj2.prop5 || obj2.prop6',
+    Obj3: 'obj3.prop1 || obj3.prop2 || obj3.prop3'
+  }
+}));
+
+describe('QueryBuilderUtils', () => {
+  describe('setQueryComputedFields', () => {
+    test('should not modify the query if there are no computed fields', () => {
+      const query = 'SELECT * FROM table WHERE field1 = "value1" AND field2 = "value2"';
```
kkerezsi, line 15: I'm confused by the usage of this SQL query. Can we change it to something that's more relevant ?

src/core/query-builder.utils.test.ts
```diff
@@ -0,0 +1,111 @@
...
+      const result = setQueryComputedFields(query);
+      expect(result).toBe(query);
+    });
+
+    test('should keep only first query of a computed field', () => {
+      const query1 = 'obj1.prop1 = "value" || obj1.prop2 = "value"';
```
kkerezsi, line 21: it might make sense to split these into separate smaller test cases.

kkerezsi:
Please add a description and a prepare a gif to highlight the functionality  in the description

src/core/query-builder.utils.test.ts
```diff
@@ -0,0 +1,111 @@
+import { expressionBuilderCallback, expressionReaderCallback, setQueryComputedFields } from "./query-builder.utils"
+
+jest.mock('./query-builder.constants.ts', () => ({
+  ...jest.requireActual('./query-builder.constants'),
+  ComputedDataFields: {
+    Obj1: 'obj1.prop1 || obj1.prop2',
```
kkerezsi, line 6: from my understanding , the agreement is not to abbreviate words .This was a feedback to me as well from @CiprianAnton

src/core/query-builder.utils.ts
```diff
@@ -0,0 +1,76 @@
+import { QueryBuilderCustomOperation } from "smart-webcomponents-react";
+import { QueryBuilderOption } from "./types";
+import { ComputedDataFields } from "./query-builder.constants";
+
+/**
+ * Removes extra queries from computed fields so the query builder will read only one
+ * @param query actual query that has been sent
```
kkerezsi, line 7: Since this is a method using a regex, it'd be great to give a small example in the description of this method.

kkerezsi (approved):

src/datasources/asset-calibration/components/AssetCalibrationQueryEditor.tsx
```diff
@@ -2,26 +2,40 @@ import { QueryEditorProps, SelectableValue, toOption } from '@grafana/data';
...
+
+  const [workspaces, setWorkspaces] = useState<Workspace[]>([]);
+
+  useEffect(() => {
+    const getWorkspaces = async () => {
+      const workspaces = await datasource.getWorkspaces();
+      setWorkspaces(workspaces);
+    }
+
+    getWorkspaces().catch(console.error);
```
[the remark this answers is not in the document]
saradei-ni: I agree, we should confirm this.

src/core/query-builder.constants.ts
```diff
@@ -0,0 +1,201 @@
...
+    timeTabLabel: 'TIME',
+    queryLabel: '',
+  },
+};
+
+export const QueryBuilderOperations = {
+  EQUALS: {
+    label: 'Equals',
+    name: '=',
+    expressionTemplate: '{0} = "{1}"',
```
CiprianAnton, line 47: You don't have to solve this now, not sure how this would work in case of numerics

src/core/query-builder.constants.ts
```diff
@@ -0,0 +1,201 @@
...
+    label: 'Contains',
+    name: 'contains',
+    expressionTemplate: '{0}.Contains("{1}")',
+  },
+  DOES_NOT_CONTAIN: {
+    label: 'Does not contain',
+    name: 'notcontains',
+    expressionTemplate: '!({0}.Contains("{1}"))',
+  },
+  IS_BLANK: {
```
CiprianAnton, line 74: These are normally not supported in query builder, do we want to support them here?

src/core/query-builder.constants.ts
```diff
@@ -0,0 +1,201 @@
...
+  IS_NOT_BLANK: {
+    label: 'Is not blank',
+    name: 'isnotblank',
+    expressionTemplate: '!string.IsNullOrEmpty({0})',
+    hideValue: true,
+  },
+  GREATER_THAN: {
+    label: 'Greater than',
+    name: '>',
+    expressionTemplate: '{0} > "{1}"',
```
CiprianAnton, line 89: Same issue related to numerics

kkerezsi (approved):

src/core/query-builder.constants.ts
saradei-ni: These were already implemented, I  just moved them in a place so we could use them more easily.

kkerezsi:
@mure  Is there any possibility to get this reviewed? The PR is out for some time now and we are expecting

src/core/types.ts
mure, line 1: I don't agree with moving all these asset types to this file. I'd be okay with `core/assets/types.ts` or  `datasources/asset-common/types.ts`, with a slight preference for the latter.

src/core/query-builder.constants.ts
```diff
@@ -0,0 +1,201 @@
...
+  QueryBuilderOperations.PROPERTY_ENDS_WITH,
+  QueryBuilderOperations.PROPERTY_CONTAINS,
+  QueryBuilderOperations.PROPERTY_DOES_NOT_CONTAIN,
+  QueryBuilderOperations.PROPERTY_IS_BLANK,
+  QueryBuilderOperations.PROPERTY_IS_NOT_BLANK,
+];
+
+export enum ComputedDataFields {
+  Location = 'Location.MinionId || Location.PhysicalLocation'
+}
```
mure, line 201: This should be a const object or an array, not an enum. I also don't think it should be in this file. `setQueryComputedFields()` should accept this object/array as an parameter and then the asset specific fields should be defined in the asset data sources.

src/datasources/asset-calibration/AssetCalibrationDataSource.ts
```diff
@@ -11,37 +11,40 @@ import {
...
   }
 
-  defaultQuery = {
-    groupBy: [],
-  };
-
   baseUrl = this.instanceSettings.url + '/niapm/v1';
 
   async runQuery(query: AssetCalibrationQuery, options: DataQueryRequest): Promise<DataFrameDTO> {
+    if (this.templateSrv.containsTemplate(query.filter)) {
```
mure, line 35: The `containsTemplate check shouldn't be necessary—if there are no variables it will just return the same string

src/datasources/asset-calibration/components/AssetCalibrationQueryEditor.tsx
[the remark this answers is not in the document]
mure: We've used this `FloatingError` component in the past as a way to display errors to the user:
https://github.com/ni/systemlink-grafana-plugins/blob/3ee9cf20081d3124b07680c0da72ab03d3d240b9/src/datasources/data-frame/components/DataFrameQueryEditor.tsx#L85

src/datasources/asset-calibration/constants.ts
```diff
@@ -0,0 +1,84 @@
...
+        filterOperations: [
+            QueryBuilderOperations.EQUALS.name,
+            QueryBuilderOperations.DOES_NOT_EQUAL.name
+        ],
+        lookup: {
+            dataSource: [
+                {
+                    label: '$workspace',
+                    value: '$workspace'
+                }
```
mure, line 26: It's been a while since I've touched the query builder. Is this giving the user an option to select a $workspace variable?

I don't think we should do this—we shouldn't assume that the user has defined a workspace variable on the dashboard or that it's called "workspace".

src/datasources/asset-calibration/constants.ts
```diff
@@ -0,0 +1,84 @@
+import { ComputedDataFields, QueryBuilderOperations } from "core/query-builder.constants";
+import { AssetTypeOptions, BusTypeOptions, QBField } from "./types";
+
+export const AssetCalibrationFields: Record<string, QBField> = {
+    LOCATION: {
+        label: 'Location',
+        dataField: ComputedDataFields.Location,
```
mure, line 7: Actually, I don't think I agree with the approach we've taken here for the "computed" fields. If I'm understanding correctly, what we're trying to do is: 
when a user inputs `Location = "foo"` in the query builder, it expands to `Location.MinionId = "foo" || Location.PhysicalLocation = "foo"`. But that expanded filter gets saved in query model, so we have to call `setQueryComputedFields()` to remove the extra filter so the query builder can parse it back to what the user expects.

Why don't we just save `Location = "foo"` in the filter, and then expand it when we make the request?

src/core/types.ts
saradei-ni: As suggested, create a separate place for those under `asset-common`

src/core/query-builder.constants.ts
saradei-ni: Created separate object that will hold required transformation, please check the new approach.

src/datasources/asset-calibration/AssetCalibrationDataSource.ts
saradei-ni: Updated as suggested, only checking for the filter to actually have values.

src/datasources/asset-calibration/constants.ts
saradei-ni: Agreed, updated the logic for this kind of query. The user can write his variable, and it will be replaced if it exists.

src/datasources/asset-calibration/constants.ts
saradei-ni: I came up with another approach that takes into consideration your suggestion. I created a function `transformComputedFieldsQuery` that takes as a parameter the query and an object that holds the template for the transformation. This will be called before sending the query.

saradei-ni:
@mure can you please take a look again? The feedback has been addressed.

src/datasources/asset-calibration/AssetCalibrationDataSource.ts
mure: I'm not even sure if that check is necessary either. The `defaultQuery` should ensure that the filter is never null/undefined.

mure (approved):

PR #73, opened 2024-10-01
feat(asset): resolve minion id

MileaRobertStefan:
# Pull Request

## 🤨 Rationale

Task [Replace SystemID](https://dev.azure.com/ni/DevCentral/_workitems/edit/2883080/).

## 👩‍💻 Implementation

On data source initialization fetch a list of systems. 

When a response of type minion ID is processed, check if the minion ID has an alias. 

## 🧪 Testing
image

Change existing tests.
Handle case where minion ID has an alias.
Handle case where minion ID does not have an alias.
Handle case where minion ID is not found.

## ✅ Checklist


kkerezsi:
Please Name the PR the right way : https://github.com/ni/systemlink-grafana-plugins#commit-message-format

src/datasources/asset-calibration/AssetCalibrationDataSource.test.ts
```diff
@@ -56,9 +58,9 @@ const locationGroupCalibrationForecastResponseMock: CalibrationForecastResponse
 {
   calibrationForecast: {
     columns: [
-      { name: "", values: [1], columnDescriptors: [{ value: "Location1", type: ColumnDescriptorType.StringValue }] },
-      { name: "", values: [2], columnDescriptors: [{ value: "Location2", type: ColumnDescriptorType.StringValue }] },
-      { name: "", values: [3], columnDescriptors: [{ value: "Location3", type: ColumnDescriptorType.StringValue }] }
+      { name: "", values: [1], columnDescriptors: [{ value: "Location1", type: ColumnDescriptorType.MinionId }] },
```
kkerezsi, line 61: Please create a new test case with MinionId.  Don't overwrite the current test because it's still a valid test. Location can be Both StringValue and MinionId.

src/datasources/asset-calibration/AssetCalibrationDataSource.ts
```diff
@@ -26,6 +28,10 @@ export class AssetCalibrationDataSource extends DataSourceBase<AssetCalibrationQ
     readonly templateSrv: TemplateSrv = getTemplateSrv()
   ) {
     super(instanceSettings, backendSrv, templateSrv);
+
+    this.querySystems().then(
+      systems => this.state.systems = systems
+    ).catch( _ => { return;});
```
kkerezsi, line 34: We should have floating errors . Please check @mure comment :

https://github.com/ni/systemlink-grafana-plugins/pull/69#discussion_r1781892686

kkerezsi:
```
 <FloatingError message={errorMsg}/> 
```

src/datasources/asset-calibration/AssetCalibrationDataSource.ts
```diff
@@ -95,7 +105,14 @@ export class AssetCalibrationDataSource extends DataSourceBase<AssetCalibrationQ
   }
 
   createColumnNameFromDescriptor(field: FieldDTOWithDescriptor): string {
-    return field.columnDescriptors.map(descriptor => descriptor.value).join(' - ');
+    return field.columnDescriptors.map(descriptor => {
+      if (descriptor.type === ColumnDescriptorType.MinionId && this.state.systems) {
```
kkerezsi, line 109: Could you have a race condition in case the systems doesn't load fast enough?

src/datasources/asset-calibration/__snapshots__/AssetCalibrationDataSource.test.ts.snap
```diff
@@ -121,7 +121,7 @@ exports[`queries calibration forecast with location groupBy 1`] = `
       {
         "name": "Group",
         "values": [
-          "Location1",
+          "my system",
```
kkerezsi, line 124: Please give it a proper name Like Minion1-alias

src/datasources/asset-calibration/AssetCalibrationDataSource.test.ts
```diff
@@ -143,6 +145,21 @@ const buildCalibrationForecastQuery = getQueryBuilder<AssetCalibrationQuery>()({
   groupBy: []
 });
 
+const fakeSystems: SystemMetadata[] = [
+  {
+    id: 'Location1',
```
kkerezsi, line 150: Minion1 instead of Location1

src/datasources/asset-calibration/AssetCalibrationDataSource.ts
kkerezsi: We also need to ensure we are falling back to default name in case this systems can't be queried

kkerezsi (changes requested):

kkerezsi:
@MileaRobertStefan you can also include a screenshot with the change.

src/datasources/asset-calibration/__snapshots__/AssetCalibrationDataSource.test.ts.snap
MileaRobertStefan: this values are used in the system test. I will change it Minion1-alias

src/datasources/asset-calibration/AssetCalibrationDataSource.ts
MileaRobertStefan: Yes it is. This is a issue I was thinking about.  I think this is an edge case.

The react whey is to use a hook that would re-render the component. But this class is not a react component.

The code I wrote will render the minion ID if the request takes to long.

We can also block the processing until the request finished. I dislike this option.

src/datasources/asset-calibration/AssetCalibrationDataSource.ts
MileaRobertStefan: I feel like I have to share the reason behind `catch( _ => { return; });`. 
I'm not sure how to mock this in test to not spam errors. 

I think that the best solution is to have a function that checks if the system is set when the runQuery is called.

MileaRobertStefan: I like this code more. Now we will query for the system when the user first runs a query. If we encounter an error then we display the IDs and we will retry to obtain the system the next time the user runs a query. 

If the system requests success we keep it in memory.

src/datasources/asset-calibration/AssetCalibrationDataSource.ts
MileaRobertStefan: Now it should be fix,

src/datasources/asset-calibration/AssetCalibrationDataSource.ts
```diff
@@ -151,6 +166,19 @@ export class AssetCalibrationDataSource extends DataSourceBase<AssetCalibrationQ
...
   }
 
+  private async loadSystems(): Promise<void> {
+    if (this.state.systems) {
+      return;
+    }
+
+    try {
+      const systems = await this.querySystems('', ['id', 'alias','connected.data.state', 'workspace']);
+      this.state.systems = new Map(systems.map(system => [system.id, system]))
```
MileaRobertStefan, line 176: I think this should be a performance boost.

src/datasources/asset-calibration/AssetCalibrationDataSource.ts
```diff
@@ -151,6 +166,19 @@ export class AssetCalibrationDataSource extends DataSourceBase<AssetCalibrationQ
     }
   }
 
+  private async loadSystems(): Promise<void> {
+    if (this.state.systems) {
+      return;
+    }
+
+    try {
+      const systems = await this.querySystems('', ['id', 'alias','connected.data.state', 'workspace']);
```
MileaRobertStefan, line 175: I see no reason to keep all the systems information. It can be quite big.

kkerezsi: Do you need the 'connected.data.state', 'workspace' ?

src/datasources/asset-calibration/AssetCalibrationDataSource.ts
```diff
@@ -151,6 +166,19 @@ export class AssetCalibrationDataSource extends DataSourceBase<AssetCalibrationQ
...
+      this.state.systems = new Map(systems.map(system => [system.id, system]))
+    } catch (error) {
+      throw new Error(`An error occurred while querying systems: ${error}`);
```
kkerezsi, line 178: Here we are losing the call stack because we are re-throwing the error with a different message that doesn't give anything valuable. I would just let the exception bubble up to have easier tracing.

src/datasources/asset-calibration/AssetCalibrationDataSource.ts
```diff
@@ -34,9 +36,15 @@ export class AssetCalibrationDataSource extends DataSourceBase<AssetCalibrationQ
 
   baseUrl = this.instanceSettings.url + '/niapm/v1';
 
+  state: AssetCalibrationDataSourceState = {
+    systems: null
+  }
```
mure, line 41: I'm not sure I see the value of the 'state' type here. I'd rather have something like:
`systemAliasCache: Map<string, SystemMetadata>`

mure: Also we may want to consider setting a TTL on the cache. Not sure how often we expect aliases to change, but we do this in the data frame plugin.
https://github.com/ni/systemlink-grafana-plugins/blob/dc1725296a6211fd49974cd56fac257b51b4a4b2/src/datasources/data-frame/DataFrameDataSource.ts#L23

mure (approved):

src/datasources/asset-calibration/AssetCalibrationDataSource.ts
MileaRobertStefan: image

these are the required fields on SystemMetadata

src/datasources/asset-calibration/AssetCalibrationDataSource.ts
MileaRobertStefan: I've read the docs on https://www.npmjs.com/package/@isaacs/ttlcache. It looks like it's great for individual items.  Given that we work with a collection I think this code will work fine.

kkerezsi (approved):

PR #77, opened 2024-10-04
feat(asset-calibration): workspace grouping

saradei-ni:
# Pull Request

## 🤨 Rationale
[User Story 2883034](https://ni.visualstudio.com/DevCentral/_workitems/edit/2883034): [Grafana][BE][UI] Add new field support for Grafana

## 👩‍💻 Implementation
Added the necessary type and mapping, also updated how workspaces are sent to the query builder.

## 🧪 Testing
Added tests for the added grouping.

image

## ✅ Checklist

### DIFF
--- a/src/datasources/asset-calibration/AssetCalibrationDataSource.test.ts
+++ b/src/datasources/asset-calibration/AssetCalibrationDataSource.test.ts
@@ -120,6 +120,38 @@ const monthLocationGroupCalibrationForecastResponseMock: CalibrationForecastResp
   }
 }
 
+const workspaceGroupCalibrationForecastResponseMock: CalibrationForecastResponse =
+{
+  calibrationForecast: {
+    columns: [
+      { name: "", values: [1], columnDescriptors: [{ value: "Workspace1", type: ColumnDescriptorType.WorkspaceId }] },
+      { name: "", values: [2], columnDescriptors: [{ value: "Workspace1", type: ColumnDescriptorType.WorkspaceId }] }
+    ]
+  }
+}
+
+const modelWorkspaceGroupCalibrationForecastResponseMock: CalibrationForecastResponse =
+{
+  calibrationForecast: {
+    columns: [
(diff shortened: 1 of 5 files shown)


src/datasources/asset-calibration/AssetCalibrationDataSource.test.ts
```diff
@@ -120,6 +120,28 @@ const monthLocationGroupCalibrationForecastResponseMock: CalibrationForecastResp
   }
 }
 
+const modelWorkspaceGroupCalibrationForecastResponseMock: CalibrationForecastResponse =
```
kkerezsi, line 123: You can probably add one more test (which will also be closer to the real-life usecase) of grouping only by workspace.

src/datasources/asset-calibration/components/AssetCalibrationQueryEditor.tsx
```diff
@@ -7,24 +7,20 @@ import { enumToOptions } from '../../../core/utils';
...
-    const getWorkspaces = async () => {
-      const workspaces = await datasource.getWorkspaces();
-      setWorkspaces(workspaces);
+    if (areWorkspacesLoaded) {
+      setWorkspaces(Array.from(datasource.workspacesCache.values()));
     }
-
-    getWorkspaces().catch(error => setError(parseErrorMessage(error) || 'Failed to fetch workspaces'));
-  }, [datasource]);
+  }, [datasource, areWorkspacesLoaded]);
```
kkerezsi, line 23: I haven't tested locally but wouldn't datasource.workspacesCache be enough/better as parameters in the dependency array of the useEffect ? This way if the datasource makes another query to the workspaces, you'd get them updated here as well rather than just checking if there are any elements in the list.

kkerezsi:
The name should be feature(asset): workspace grouping

src/datasources/asset-calibration/components/AssetCalibrationQueryEditor.tsx
saradei-ni: Changed the approach here due to recent clarifications.

kkerezsi (approved):

PR #81, opened 2024-10-11
feat(asset): load dependencies only once, change ttlcache to map

saradei-ni:
# Pull Request

We don't need Cache for our dependencies, storing them in memory once it's enough.
Setting the dependencies in queryEditor wasn't consistent.
Dependencies wouldn't load unless a query was made.

## 👩‍💻 Implementation

Used Map for easy retrieval of data instead of TTLCache.
Updated how the queryEditor knows when the loading is finished, used rxjs for it in order to make sure we load the data after everything is set.
Dependencies are loading regardless, if a query is ran at the start it will wait for the dependencies to load, no extra loading is done.

## 🧪 Testing
Plugin behaves the same.

## ✅ Checklist

- [x ] This PR has a title that follows the [commit message format](https://github.com/ni/systemlink-grafana-plugins#commit-message-format).

### DIFF
--- a/src/datasources/asset-calibration/AssetCalibrationDataSource.ts
+++ b/src/datasources/asset-calibration/AssetCalibrationDataSource.ts
@@ -23,13 +23,12 @@ import {
 import { ExpressionTransformFunction, transformComputedFieldsQuery } from 'core/query-builder.utils';
 import { AssetCalibrationFieldNames } from './constants';
 import { AssetModel, AssetsResponse } from 'datasources/asset-common/types';
-import TTLCache from '@isaacs/ttlcache';
-import { metadataCacheTTL } from 'datasources/data-frame/constants';
 import { SystemMetadata } from 'datasources/system/types';
 import { defaultOrderBy, defaultProjection } from 'datasources/system/constants';
 import { QueryBuilderOperations } from 'core/query-builder.constants';
 import { Workspace } from 'core/types';
 import { parseErrorMessage } from 'core/errors';
+import { Subject } from 'rxjs';
 
 export class AssetCalibrationDataSource extends DataSourceBase<AssetCalibrationQuery, DataSourceJsonData> {
   public defaultQuery = {
@@ -37,13 +36,13 @@ export class AssetCalibrationDataSource extends DataSourceBase<AssetCalibrationQ
     filter: ''
   };
 
-  public areSystemsLoaded = false;
-  public areWorkspacesLoaded = false;
+  public areSystemsLoaded$ = new Subject<void>();
+  public areWorkspacesLoaded$ = new Subject<void>();
 
   public error = '';
 
-  public readonly systemAliasCache: TTLCache<string, SystemMetadata> = new TTLCache<string, SystemMetadata>({ ttl: metadataCacheTTL });
-  public readonly workspacesCache: TTLCache<string, Workspace> = new TTLCache<string, Workspace>({ ttl: metadataCacheTTL });
+  public readonly systemAliasCache = new Map<string, SystemMetadata>([]);
+  public readonly workspacesCache = new Map<string, Workspace>([]);
   public readonly busTypeCache = new Map<BusType, string>([
     ...BusTypeOptions.map(busType => [busType.value, busType.label]) as Array<[BusType, string]>
   ]);
@@ -52,6 +51,7 @@ export class AssetCalibrationDataSource extends DataSourceBase<AssetCalibrationQ
   ]);
 
   private readonly baseUrl = this.instanceSettings.url + '/niapm/v1';
+  private dependenciesLoadedPromise: Promise<void>;
 
   constructor(
     readonly instanceSettings: DataSourceInstanceSettings,
@@ -59,10 +59,14 @@ export class AssetCalibrationDataSource extends DataSourceBase<AssetCalibrationQ
     readonly templateSrv: TemplateSrv = getTemplateSrv(),
   ) {
     super(instanceSettings, backendSrv, templateSrv);
+    this.dependenciesLoadedPromise = this.loadDependencies();
   }
 
   async runQuery(query: AssetCalibrationQuery, options: DataQueryRequest): Promise<DataFrameDTO> {
-    await this.loadDependencies();
+    await this.dependenciesLoadedPromise;
+
+    this.error = '';
+
     this.validateQueryRange(query, options);
 
     if (query.filter) {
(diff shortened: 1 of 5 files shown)


src/datasources/asset-calibration/AssetCalibrationDataSource.ts
```diff
@@ -18,42 +18,43 @@ import {
...
 import { AssetCalibrationFieldNames } from './constants';
 import { AssetModel, AssetsResponse } from 'datasources/asset-common/types';
-import TTLCache from '@isaacs/ttlcache';
-import { metadataCacheTTL } from 'datasources/data-frame/constants';
 import { SystemMetadata } from 'datasources/system/types';
 import { defaultOrderBy, defaultProjection } from 'datasources/system/constants';
 import { QueryBuilderOperations } from 'core/query-builder.constants';
 import { Workspace } from 'core/types';
 import { parseErrorMessage } from 'core/errors';
+import { Subject } from 'rxjs';
```
kkerezsi, line 26: @saradei-ni I don't think we should use rxjs with React. 

RxJS is not an idiomatic way to manage ui state changes within React

kkerezsi: Please use hooks instead. If needed, you can also implement a custom hook and trigger rendering when the map is updated

kkerezsi: u can also add a useState hook and useRef hooks to your maps to avoid rerendering on change of the map and trigger only when the entire dependency set is loaded

saradei-ni: I have updated the approach to use promises instead of rxjs. The resolve will be triggered after loading is complete. The editor can set the state afterwards knowing the loading has been done, it waits for both promises to resolve.

This reactive approach ensures the editor is notified when the loading is complete. The previous approach wasn't consistent as the datasource object is only loaded once, and the async changes wouldn't properly reflect in editor when the flags were changed. Also, this covers the case when a query is not made at init, and the dependencies wouldn't load.

Let me know if we need any additional changes.

saradei-ni:
@mure we needed a small type update in the core, can you please check it as well?

kkerezsi (approved):

src/datasources/asset-calibration/AssetCalibrationDataSource.ts
```diff
@@ -37,13 +35,16 @@ export class AssetCalibrationDataSource extends DataSourceBase<AssetCalibrationQ
...
-  public areWorkspacesLoaded = false;
+  private systemsLoaded!: () => void;
+  private workspacesLeaded!: () => void;
+
+  public areSystemsLoaded$ = new Promise<void>(resolve => this.systemsLoaded = resolve);
+  public areWorkspacesLoaded$ = new Promise<void>(resolve => this.workspacesLeaded = resolve);
 
   public error = '';
 
-  public readonly systemAliasCache: TTLCache<string, SystemMetadata> = new TTLCache<string, SystemMetadata>({ ttl: metadataCacheTTL });
```
MileaRobertStefan, line 45: @saradei-ni are we sure we don't want to invalidate the cache?

Catalin-web (approved):

src/datasources/asset-calibration/__snapshots__/AssetCalibrationDataSource.test.ts.snap
```diff
@@ -195,7 +195,7 @@ exports[`queries calibration forecast with minion ID location groupBy 1`] = `
       {
         "name": "Group",
         "values": [
-          "Minion1-alias",
```
MileaRobertStefan, line 198: What happened to the alias?

MileaRobertStefan (approved):

src/datasources/asset-calibration/AssetCalibrationDataSource.ts
```diff
@@ -52,17 +53,22 @@ export class AssetCalibrationDataSource extends DataSourceBase<AssetCalibrationQ
   ]);
 
   private readonly baseUrl = this.instanceSettings.url + '/niapm/v1';
+  private dependenciesLoadedPromise: Promise<void>;
```
mure, line 56: I think we could approach this in a way that keeps us from having to manage all this state. The caches should be implementation details that aren't exposed to "clients" (the query editor and other parts of the data source), and they should be lazy. We already have a `getWorkspaces()` method that returns cached data: https://github.com/ni/systemlink-grafana-plugins/blob/ac08a7d893e727beff1135c90a55b3502a906c2f/src/core/DataSourceBase.ts#L65

Then we can write a `async getSystemAliasMap()` method that does the same thing—makes a request the first time it's called and returns cached data after that. Then anytime we need this data, we just call those methods. In the query editor, use the `useAsync` hook for a clean way to manage the loading state: 

https://github.com/ni/systemlink-grafana-plugins/blob/8a028a82a2688d105eb2e165fcf83a558e2069ba/src/datasources/data-frame/components/DataFrameQueryEditor.tsx#L18

saradei-ni: I am not sure that the implemented cache is working as expected, take a look at this for example.

image

We log the instance, workspaces, and a counter to track how many times the function is hit.
We take a dashboard that loads workspace-datasource 2 times and asset-calibration one time.  The setup produces the following results.

image

We can notice that the state is shared only between same type datasource, so even if we keep this kind of cache, it will only be accessible to same type datasources. Even if this is what we want, current implementation is not really working, if the function is called too fast it will trigger multiple calls, not waiting for a stream of data that is already started.

We should take into consideration a solution that can share necessary data between different sources, maybe even invalidate and refresh the data as well. The way plugins are managed by grafana doesn't seem to allow us to do this by storing in memory data in a base class. We should talk about alternative ways to store this type of data so we can get a result like this one for example.

image

Taking this into consideration I don't really see a benefit in moving this logic in base, at least for now.

mure: @saradei-ni and I discussed this in a call. 

We agreed that the cache for `DataSourceBase.getWorkspaces()` has race conditions—if there are multiple callers before the fetch returns, we will make duplicate requests. I will fix this in a separate change and unblock this PR. In a later change we can apply the same approach for caching systems. 

There's another issue which is that plugin modules are isolated from each other, so there's not a straightforward way to share data between different plugins. So if you have 3 different data sources on a dashboard that all need workspace metadata, we would make 3 separate requests. Resolving this will take some research.

mure (approved):

PR #96, opened 2024-10-21
feat(asset): added support for IncludeOnlyDataInTimeRange

saradei-ni:
# Pull Request

## 🤨 Rationale

We need the ability to add a flag so the service knows that it needs to query only data in the interval.

## 👩‍💻 Implementation

Updated data links and runQuery method to take into consideration IncludeOnlyDataInTimeRange

## 🧪 Testing

Added tests for the flag

## ✅ Checklist

### DIFF
--- a/src/datasources/asset/data-sources/AssetDataSourceBase.ts
+++ b/src/datasources/asset/data-sources/AssetDataSourceBase.ts
@@ -45,6 +45,14 @@ export abstract class AssetDataSourceBase extends DataSourceBase<AssetQuery, Ass
     }
   }
 
+  getQueryParams(): URLSearchParams {
+    return new URLSearchParams(window.location.search);
+  }
+
+  getQueryParam(param: string): string | null {
+    return this.getQueryParams().get(param);
+  }
+
   public getCachedSystems(): SystemMetadata[] {
     return Array.from(this.systemAliasCache.values());
   }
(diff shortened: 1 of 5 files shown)


src/datasources/asset/data-sources/AssetDataSourceBase.ts
```diff
@@ -45,6 +45,14 @@ export abstract class AssetDataSourceBase extends DataSourceBase<AssetQuery, Ass
     }
   }
 
+  getQueryParams(): URLSearchParams {
+    return new URLSearchParams(window.location.search);
```
kkerezsi, line 49: @saradei-ni I'd double check that grafana uses the same mechanism in their sourcecode in to get the variables. My only concern is to have a usecase where this code is not running in the browser.  Maybe it would be better to pass the query params as parameters to the datasource 

https://grafana.com/developers/plugin-tools/tutorials/build-a-data-source-backend-plugin

kkerezsi (approved):

kkerezsi:
One more thing I find strange is that he user doesn't know when they are with timerange only selected...

We should probably have some sort of possibility to show that this is active (maybe via a global variable) and not render the datalinks at all when you are with timerange only active

src/datasources/asset/data-sources/AssetDataSourceBase.ts
saradei-ni: the rest of variables are handled via query params as well
I added a way to handle already in time range case as well.

PR #110, opened 2024-11-05
fix(notebook): fix input sanitization for notebook query builder

kkerezsi:
# Pull Request

## 🤨 Rationale

Since the query builder is vulnerable to XSS, we need to ensure that the customers are safe using this feature.

## 👩‍💻 Implementation

Added sanitization to the filter

## 🧪 Testing

- Unit testing
- Manual testing

## ✅ Checklist

### DIFF
--- a/src/datasources/notebook/QueryBuilder.tsx
+++ b/src/datasources/notebook/QueryBuilder.tsx
@@ -8,6 +8,7 @@ import 'smart-webcomponents-react/source/styles/smart.orange.css';
 import 'smart-webcomponents-react/source/styles/components/smart.base.css';
 import 'smart-webcomponents-react/source/styles/components/smart.common.css';
 import 'smart-webcomponents-react/source/styles/components/smart.querybuilder.css';
+import { filterXSSLINQExpression } from 'core/utils';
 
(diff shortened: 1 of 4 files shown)


Catalin-web (approved):

src/datasources/notebook/QueryBuilder.tsx
```diff
@@ -162,8 +163,9 @@ export const TestResultsQueryBuilder: React.FC<TestResultsQueryBuilderProps> = (
       messages={queryBuilderMessages}
       showIcons
       // Only set value on first render
-      {...(initialize.current && { value: props.defaultValue })}
+      {...(initialize.current && { value: filterXSSLINQExpression(props.defaultValue) })}
       {...props}
+      value={filterXSSLINQExpression(props.defaultValue)}
```
kkerezsi, line 168: @mure Is there a reason why you haven't been using the `value` attribute in the query builder and instead you were only using the `defaultValue` ?

mure: Are you just asking why I named the prop `defaultValue` and not `value`?

I can't remember exactly why, it's been a long time since I touched this code. Probably because it's a convention in React that `defaultValue` sets the initial value for the control. `value` is used for controlled components: https://legacy.reactjs.org/docs/forms.html#controlled-components

mure (approved):

PR #111, opened 2024-11-14
fix(asset): calibration forecast timezone issues on datalinks

kkerezsi:
# Pull Request

## 🤨 Rationale

- Due to a reported issue in the CST time zone, it seems that the calibration forecast dashboard doesn't build the datalinks properly
- The bucket label was also built in local time instead of UTC

## 👩‍💻 Implementation

- The datalink time building algorithm was adjusted to account for the actual values from the backend when building the links
- The Buckets labels respect the UTC values now

## 🧪 Testing

- Unit testing
- Manual testing on different time zones

## ✅ Checklist

### DIFF
--- a/src/datasources/asset/data-sources/calibration-forecast/CalibrationForecastDataSource.ts
+++ b/src/datasources/asset/data-sources/calibration-forecast/CalibrationForecastDataSource.ts
@@ -66,8 +66,8 @@ export class CalibrationForecastDataSource extends AssetDataSourceBase {
 
     async processCalibrationForecastQuery(query: CalibrationForecastQuery, options: DataQueryRequest) {
         const result: DataFrameDTO = { refId: query.refId, fields: [] };
-        const from = options.range!.from.toISOString();
-        const to = options.range!.to.toISOString();
+        const from = options.range!.from.utc().toISOString();
+        const to = options.range!.to.utc().toISOString();
 
         const calibrationForecastResponse: CalibrationForecastResponse = await this.queryCalibrationForecast(
             query.groupBy,
@@ -102,6 +102,19 @@ export class CalibrationForecastDataSource extends AssetDataSourceBase {
     }
 
     processResultsGroupedByTime(result: DataFrameDTO, timeGrouping: AssetCalibrationForecastKey) {
+        const originalValuesMap = new Map<string, string>();
+        if (result.fields.length > 0) {
+            const field = result.fields[0];
+            const originalValues = field.values as string[] ?? [];
+            const formatter = this.forecastDateFormatterMap.get(field.name as AssetCalibrationForecastKey);
+            if (formatter) {
+                const formattedValues: string[] = field.values!.map(formatter);
+                for (let i = 0; i < originalValues!.length; i++) {
+                    originalValuesMap.set(formattedValues![i], originalValues![i]);
+                }
+            }
+        }
+
         result.fields.forEach(field => {
             field.name = this.createColumnNameFromDescriptor(field as FieldDTOWithDescriptor);
             const formatter = this.forecastDateFormatterMap.get(field.name as AssetCalibrationForecastKey);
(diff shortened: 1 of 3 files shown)


src/datasources/asset/data-sources/calibration-forecast/CalibrationForecastDataSource.ts
```diff
@@ -101,7 +101,8 @@ export class CalibrationForecastDataSource extends AssetDataSourceBase {
         return null;
     }
 
-    processResultsGroupedByTime(result: DataFrameDTO, timeGrouping: AssetCalibrationForecastKey) {
+    processResultsGroupedByTime(result: DataFrameDTO, timeGrouping: AssetCalibrationForecastKey, options: DataQueryRequest) {
+        const originalValuesMap = this.constructDataLinkValuesMap(result, options);
```
saradei-ni, line 105: As the method name is `constructDataLinkValuesMap`, maybe `dataLinkValuesMap` would be more suitable name.

src/datasources/asset/data-sources/calibration-forecast/CalibrationForecastDataSource.ts
```diff
@@ -101,7 +101,8 @@ export class CalibrationForecastDataSource extends AssetDataSourceBase {
...
         result.fields.forEach(field => {
             field.name = this.createColumnNameFromDescriptor(field as FieldDTOWithDescriptor);
             const formatter = this.forecastDateFormatterMap.get(field.name as AssetCalibrationForecastKey);
```
saradei-ni, line 108: While here, `dateFormatter` might be a better name for this as well.

src/datasources/asset/data-sources/calibration-forecast/CalibrationForecastDataSource.ts
```diff
@@ -111,17 +112,58 @@ export class CalibrationForecastDataSource extends AssetDataSourceBase {
...
             if (!formatter) {
-                field.config = { links: this.createDataLinks(timeGrouping) };
+                field.config = { links: this.createDataLinks(timeGrouping, originalValuesMap) };
             }
         });
     }
 
-    private constructDataLinkBaseUrl(){
+    private constructDataLinkValuesMap(result: DataFrameDTO, options: DataQueryRequest): Map<string, { from: number, to: number }> {
+        const originalValuesMap = new Map<string, { from: number, to: number }>();
```
saradei-ni, line 121: `dataLinkValuesMap` might be more suited here as well.

src/datasources/asset/data-sources/calibration-forecast/CalibrationForecastDataSource.ts
```diff
@@ -111,17 +112,58 @@ export class CalibrationForecastDataSource extends AssetDataSourceBase {
...
+    }
+
+    private constructDataLinkBaseUrl() {
         const pathname = locationService.getLocation().pathname;
-        return  pathname + '?orgId=${__org.id}&${__all_variables}';
+        return pathname + '?orgId=${__org.id}&${__all_variables}';
     }
 
-    private createDataLinks(timeGrouping: AssetCalibrationForecastKey): DataLink[] {
+    private createDataLinks(timeGrouping: AssetCalibrationForecastKey, originalValuesMap: Map<string, { from: any, to: any }>): DataLink[] {
```
saradei-ni, line 166: The I think is `{ from: number, to: number }`, so we can avoid any.

src/datasources/asset/data-sources/calibration-forecast/CalibrationForecastDataSource.ts
```diff
@@ -130,36 +172,12 @@ export class CalibrationForecastDataSource extends AssetDataSourceBase {
                     if (!options.replaceVariables) {
                         return url;
                     }
-
                     const value = options.replaceVariables(`\${__data.fields.${timeGrouping}}`);
```
saradei-ni, line 175: We could have something like this if we don't have a value for the map every time.

```
                    const value = options.replaceVariables(`\${__data.fields.${timeGrouping}}`);
                    const valueAsDate = originalValuesMap.get(value);
                    if (!value && !valueAsDate) {
                        return url;
                    }

                    return `${url}&from=${valueAsDate.from}&to=${valueAsDate.to}`;
```

If we do have a value every time, what about something like this?

```
                    const value = options.replaceVariables(`\${__data.fields.${timeGrouping}}`);
                    if (!value) {
                        return url;
                    }

                    const [from, to] = originalValuesMap.get(value)!;
                    return `${url}&from=${from}&to=${to}`;
```

PR #121, opened 2025-02-05
feat(results): Add floating error in results query editor

Ahalya-ni:
# Pull Request

## 🤨 Rationale
Feature : [Feature 2605755 ](https://dev.azure.com/ni/DevCentral/_workitems/edit/2605755): Grafana support for results datasource

As a part of [User Story 2980051](https://dev.azure.com/ni/DevCentral/_workitems/edit/2980051): Add Results Query editors to the Results Datasource, this PR contains the addition of `FloatingError` reusable for handling errors in Results Query Editor. 

## 👩‍💻 Implementation

- The errors are parsed using `parseErrorMessage` from `core/error`.
- The parsed error message is passed to the `FloatingError` component to be displayed in the Results Query Editor when errors occur.

## 🧪 Testing

Updated test

## ✅ Checklist

### DIFF
--- a/src/datasources/results/ResultsDataSource.test.ts
+++ b/src/datasources/results/ResultsDataSource.test.ts
@@ -5,7 +5,7 @@ import { BackendSrv, getTemplateSrv } from '@grafana/runtime';
 import { createFetchError, createFetchResponse, getQueryBuilder, requestMatching, setupDataSource } from 'test/fixtures';
 import { Field } from '@grafana/data';
 
-const mockQueryResulltsResponse: QueryResultsResponse = {
+const mockQueryResultsResponse: QueryResultsResponse = {
   results: [
     {
       id: '000007fb-aa87-4ab9-9757-6568e7893c33',
@@ -18,7 +18,8 @@ const mockQueryResulltsResponse: QueryResultsResponse = {
 };
 
 jest.mock('@grafana/runtime', () => ({
-  getTemplateSrv: jest.fn()
+  getTemplateSrv: jest.fn(),
+  isFetchError: jest.fn()
 }));
 
 let datastore: ResultsDataSource, backendServer: MockProxy<BackendSrv>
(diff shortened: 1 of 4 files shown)


src/datasources/results/ResultsDataSource.test.ts
```diff
@@ -18,7 +18,8 @@ const mockQueryResulltsResponse: QueryResultsResponse = {
 };
 
 jest.mock('@grafana/runtime', () => ({
-  getTemplateSrv: jest.fn()
+  getTemplateSrv: jest.fn(),
+  isFetchError: jest.fn()
```
richie-ni, line 22: Do we really need to mock it? Is there a way to work this up just with the backendSrv

src/datasources/results/ResultsDataSource.test.ts
```diff
@@ -27,13 +28,14 @@ describe('ResultsDataSource', () => {
   beforeEach(() => {
     (getTemplateSrv as jest.Mock).mockReturnValue({
```
richie-ni, line 29: I hope we can use the datastore.getTemplateSrv to get the replace method

richie-ni, line 29: Can we also write a test case to check if the filter contains the templatesrv replaced value

richie-ni (changes requested):

src/datasources/results/ResultsDataSource.test.ts
Ahalya-ni: It is covered in the base datasource PR for result.

src/datasources/results/ResultsDataSource.test.ts
Ahalya-ni: It is not needed. Please refer to the `QueryResultsDataSource` test. It is covered in other PR: Create base datasource for results

Ahalya-ni:
As the reusable `FloatingError` has issues with the component not being destroyed, Richie Soshan worked on creating a Toast message for Grafana Datasources. This work is currently under review. For now, it has decided to handle the error in the same way as the Product datasource, which has already been applied to the Results datasource. Hence abandoning this PR.

PR #130, opened 2025-02-20
fix(products): Prevent non-numeric input in record count field

richie-ni:
# Pull Request

## 🤨 Rationale
On testing the Products Datasource on different browsers, we noticed that the `autoSizeInput` element with Number type used for `Take` field which should allow only Numbers, works properly in `Edge` but allows Alphabets and symbols in `Firefox` and `Chrome` . 

image

This PR has the changes to enhance input validation for the Take field in all browsers.

## 👩‍💻 Implementation
-   Added a `checkIfNumber` function to ensure that only numeric values or specific control keys (Backspace, Tab, Delete, ArrowLeft, ArrowRight) are allowed in the record count input field.

- Integrated the `checkIfNumber` function with the `onKeyDown` event of the record count input field to enforce the validation.

## 🧪 Testing
Tested manually in all browsers
Added unit tests to verify the same

## ✅ Checklist

- [ x ] This PR has a title that follows the [commit message format](https://github.com/ni/systemlink-grafana-plugins#commit-message-format).

### DIFF
--- a/src/datasources/products/components/ProductsQueryEditor.tsx
+++ b/src/datasources/products/components/ProductsQueryEditor.tsx
@@ -69,6 +69,11 @@ export function ProductsQueryEditor({ query, onChange, onRunQuery, datasource }:
     }
   }
 
+  function checkIfNumber(event: React.KeyboardEvent<HTMLInputElement>) {
+    const regex = new RegExp(/(^-?\d*$)|(Backspace|Tab|Delete|ArrowLeft|ArrowRight)/);
+    return !event.key.match(regex) && event.preventDefault();
+ }
+
   return (
     <>
       <HorizontalGroup align="flex-start">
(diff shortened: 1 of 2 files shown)


ranjgith-ni (approved):

src/datasources/products/components/ProductsQueryEditor.tsx
```diff
@@ -69,6 +69,11 @@ export function ProductsQueryEditor({ query, onChange, onRunQuery, datasource }:
     }
   }
 
+  function checkIfNumber(event: React.KeyboardEvent<HTMLInputElement>) {
+    const regex = new RegExp(/(^-?\d*$)|(Backspace|Tab|Delete|ArrowLeft|ArrowRight)/);
```
cameronwaterman, line 73: Would it be better to parse to a float and then check for NaN instead?
Maybe something like this:

`!isNaN(value) && !isNaN(parseFloat(value))`

cameronwaterman (approved):

src/datasources/products/components/ProductsQueryEditor.tsx
richie-ni: This would work for Chrome but not in Firefox , Firefox still allows non-numeric value in the input field if we parse it to a float and then check for a NaN.

PR #138, opened 2025-04-08
refactor(products): move key-value operation utilities to core

richie-ni:
# Pull Request

## 🤨 Rationale

As part of the Product Datasource, we introduced a custom property called Key-Value, along with its custom template and expressions. To enable the use of Key-Value operations in other datasource query builders, we need to move the related operation utilities out of their current location and into the core component for broader accessibility and reuse.

## 👩‍💻 Implementation

- Moved the `Key Value Operation Utilities` into Core folder under `customOperations` folder
- Renamed the files to KeyValueOperations

## 🧪 Testing

N/A 

## ✅ Checklist

- [ x  ] This PR has a title that follows the [commit message format](https://github.com/ni/systemlink-grafana-plugins#commit-message-format).


src/core/customQueryBuilderOperations/keyValueOperations.test.ts
```diff
@@ -1,7 +1,7 @@
 import { FilterOperations } from "core/query-builder.constants";
```
rajsite, line 1: We don't have perfect consistency but at a high-level it looks like we have the naming convention of:
- folders: /lower-snake-case
- components / classes: PascalCaseBasedOnComponentOrClass.ts
- utility files: lower-snake-case.ts

Could we align with those?
```
src/core/custom-query-builder-operations/key-value-operations.ts
src/core/custom-query-builder-operations/key-value-operations.test.ts
```

rajsite (approved):

src/core/customQueryBuilderOperations/keyValueOperations.test.ts
richie-ni: Changed as suggested

cameronwaterman (approved):

PR #143, opened 2025-04-15
fix(core): Update numeric input validation method to support negative values

Ahalya-ni:
# Pull Request

## 🤨 Rationale
As a part of this [User Story 2980051](https://ni.visualstudio.com/DevCentral/_workitems/edit/2980051): Add Results Query editors to the Results Datasource

This PR updates  `validateNumericInput` method to allow the `-` input, enabling for negative values.

## 👩‍💻 Implementation
- Modified the `validateNumericInput` method in the `utils.ts` file to include the `-` sign as a valid input.

## 🧪 Testing

- Added unit test. 

## ✅ Checklist

### DIFF
--- a/src/core/utils.ts
+++ b/src/core/utils.ts
@@ -112,7 +112,7 @@ export function filterXSSLINQExpression(value: string | null | undefined): strin
 }
 
 export function validateNumericInput(event: React.KeyboardEvent<HTMLInputElement>) {
-  if (isNaN(Number(event.key)) && !['Backspace', 'Tab', 'Delete', 'ArrowLeft', 'ArrowRight'].includes(event.key)) {
+  if (isNaN(Number(event.key)) && !['Backspace', 'Tab', 'Delete', 'ArrowLeft', 'ArrowRight', '-'].includes(event.key)) {
     event.preventDefault();
   }
 }
--- a/src/core/utils.test.ts
+++ b/src/core/utils.test.ts
@@ -62,6 +62,14 @@ describe('validateNumericInput', () => {
 
       expect(mockPreventDefault).not.toHaveBeenCalled();
     });
+
+    test('allows negative sign', () => {
+        const event = { key: '-', preventDefault: mockPreventDefault } as unknown as React.KeyboardEvent<HTMLInputElement>;
+        
+        validateNumericInput(event);
+  
+        expect(mockPreventDefault).not.toHaveBeenCalled();
+    });
   
     test('allows navigation keys', () => {
         const event = { key: 'Tab', preventDefault: mockPreventDefault } as unknown as React.KeyboardEvent<HTMLInputElement>;


src/core/utils.ts
```diff
@@ -112,7 +112,7 @@ export function filterXSSLINQExpression(value: string | null | undefined): strin
 }
 
 export function validateNumericInput(event: React.KeyboardEvent<HTMLInputElement>) {
-  if (isNaN(Number(event.key)) && !['Backspace', 'Tab', 'Delete', 'ArrowLeft', 'ArrowRight'].includes(event.key)) {
+  if (isNaN(Number(event.key)) && !['Backspace', 'Tab', 'Delete', 'ArrowLeft', 'ArrowRight', '-'].includes(event.key)) {
```
ranjgith-ni, line 115: This method will be used in validating the `take` input. I am unclear on why we would want a negative `take`

ranjgith-ni (changes requested):

src/core/utils.ts
richie-ni: We wanted to allow the negative sign in the take input because the API uses -1 to indicate that all results should be fetched for a given query.

ranjgith-ni (approved):

src/core/utils.ts
```diff
@@ -112,7 +112,7 @@ export function filterXSSLINQExpression(value: string | null | undefined): strin
 }
 
 export function validateNumericInput(event: React.KeyboardEvent<HTMLInputElement>) {
```
atmgrifter00, line 114: How else do we expect this API to be used? Is this also valid for fractional values (i.e. do we need to accept '.' as a valid character)? Where else do we provide validation for the complete value, as we can't, for example, accept a value of '-1-2'?

Ahalya-ni: The API allows only integer values. Yes, we do provide [complete value validation](https://github.com/ni/systemlink-grafana-plugins/blob/b29f0b7c50fdd49d1fca5ba2ecaaf727680896ad/src/datasources/results/components/editors/query-results/QueryResultsEditor.tsx#L43) in the `onCommitChange` event. For cases where invalid inputs like `-1-1` is entered, they are parsed to an integer using the `parseInt` method.

atmgrifter00: So, calling `parseInt` on an improperly formatted value doesn't guarantee a valid value. For example "--1-2" results in `NaN`. The remaining question is whether or not the APIs/UI are robust enough to handle `NaN`.

rajsite, line 114: Taking one step back, it looks like validateNumericInput is only being used on different AutoSizeInputs [[ref1]](https://github.com/ni/systemlink-grafana-plugins/blob/b29f0b7c50fdd49d1fca5ba2ecaaf727680896ad/src/datasources/products/components/ProductsQueryEditor.tsx#L131) [[ref2]](https://github.com/ni/systemlink-grafana-plugins/blob/b29f0b7c50fdd49d1fca5ba2ecaaf727680896ad/src/datasources/results/components/editors/query-results/QueryResultsEditor.tsx#L99)

AutoSizeInput also is just [forwarding props to a native html input](https://github.com/grafana/grafana/blob/d1b2e669a70ac73a28f9089f0afa124aaa31cff8/packages/grafana-ui/src/components/Input/AutoSizeInput.tsx#L54).

Instead of trying to intercept keyboard handling and trying to pass through all these special cases with arrow keys, etc, maybe we can just specify a [`pattern`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/pattern).

That would be a much simpler and expressive way to control what is allowed to be entered. Example [jsbin](https://jsbin.com/gucoxehovi/edit?html,console,output).

rajsite: (this also implies using checkValidity in the input and reporting errors, etc, which is a different UX, but less magical)

richie-ni: @rajsite

The main reason for adding the validator to the `AutoSizeInput` was to ensure consistent behavior across browsers when restricting non-numeric input. Setting just the input type to number doesn’t fully prevent users from entering non-numeric characters—browsers like Firefox and Chrome still allow alphabetic characters, while Edge properly restricts them. 

Similarly, using a pattern attribute works as expected in Edge, but fails to consistently block non-numeric input in other browsers for the Take input - 
(In Firefox )
image

rajsite: > ensure consistent behavior across browsers when restricting non-numeric input. 

> this also implies using checkValidity in the input and reporting errors, etc, which is a different UX

The change in UX I'm describing is to avoid restricting "typing" non-numeric input (there are other ways to enter text like copy/paste, international keyboards, etc that validating input events is incomplete / hacky) and instead to validate the content of the input and only accept it when it is valid.

richie-ni:
Following the recent discussion on allowing negative integers in the take input, it has been decided not to support negative values. As a result, the changes proposed in the PR are no longer applicable. I will go ahead and close this PR and will plan to create a newer PR with the suggestion of Milan.

PR #148, opened 2025-04-17
refactor(tags): individual classes for handling different types of queries

saradei-ni:
# Pull Request

## 🤨 Rationale

[User Story 3036390](https://ni.visualstudio.com/DevCentral/_workitems/edit/3036390): [Grafana] Revisit code & enable multiselect values for tags plugin
[Task 3086953](https://ni.visualstudio.com/DevCentral/_workitems/edit/3086953): Update datasource code, make it easier to understand

## 👩‍💻 Implementation

Created individual files for handling different types of queries (current and historical). Broke down historical query handler into separate methods that describe the logic, adjusted the order and logic.

## 🧪 Testing
All tests pass

## ✅ Checklist

### DIFF
--- a/src/datasources/tag/CurrentQueryHandler.ts
+++ b/src/datasources/tag/CurrentQueryHandler.ts
@@ -9,7 +9,16 @@ export class CurrentQueryHandler extends QueryHandler {
     }
 
     private handleCurrentQuery(queryProperties: boolean, tagsWithValues: TagWithValue[], result: DataFrameDTO): DataFrameDTO {
-        const allPossibleProps = this.getAllProperties(tagsWithValues);
+        this.addDefaultFieldsToResult(result, tagsWithValues);
+
+        if (queryProperties) {
+            this.addPropertiesFieldsToResult(result, tagsWithValues);
+        }
+
+        return result;
+    }
+
+    private addDefaultFieldsToResult(result: DataFrameDTO, tagsWithValues: TagWithValue[]): void {
         result.fields = [
             {
                 name: 'name',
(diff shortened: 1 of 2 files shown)


src/datasources/tag/utils.ts
```diff
@@ -9,6 +9,10 @@ export function expandMultipleValueVariable(input: string): string[] {
   return results;
 }
 
+export function convertTagValue(type: string, value?: string) {
+  return value && ['DOUBLE', 'INT', 'U_INT64'].includes(type) ? Number(value) : value;
```
kkerezsi, line 13: we could cache `['DOUBLE', 'INT', 'U_INT64']` and just reuse it to avoid having this array created on each call.

src/datasources/tag/HistoricalQueryHandler.ts
```diff
@@ -0,0 +1,217 @@
...
+    private async getTagHistoryWithChunks(paths: TagWithValue[], workspace: string, range: TimeRange, intervals: number): Promise<TagHistoryResponse> {
+        const chunkSize = 10;
+        const pathChunks: TagWithValue[][] = [];
+        for (let i = 0; i < paths.length; i += chunkSize) {
+            pathChunks.push(paths.slice(i, i + chunkSize));
+        }
+
+        const aggregatedResults: TagHistoryResponse = { results: {} };
+
+        // Fetch and aggregate the data from each chunk in parallel
```
kkerezsi, line 118: We can remove the comments here and below

kkerezsi (approved):

src/datasources/tag/utils.ts
saradei-ni: Moved the array into a separate constant inside the file.

src/datasources/tag/HistoricalQueryHandler.ts
saradei-ni: Removed comments.

src/datasources/tag/CurrentQueryHandler.ts
```diff
@@ -0,0 +1,59 @@
...
+import { Workspace } from "core/types";
+
+export class CurrentQueryHandler extends QueryHandler {
+    handleQuery(tagsWithValues: TagWithValue[], result: DataFrameDTO, _workspaces: Workspace[], _range: TimeRange, _maxDataPoints: number | undefined, queryProperties: boolean): Promise<DataFrameDTO> {
+        return Promise.resolve(this.handleCurrentQuery(queryProperties, tagsWithValues, result));
+    }
+
+    private handleCurrentQuery(queryProperties: boolean, tagsWithValues: TagWithValue[], result: DataFrameDTO): DataFrameDTO {
+        const allPossibleProps = this.getAllProperties(tagsWithValues);
+        result.fields = [
```
atmgrifter00, line 13: I like what was done in `HistoricalQueryHandler` with the private methods for updating the `result`. Can we follow that pattern here? I think it would make it more readable/understandable.

src/datasources/tag/utils.ts
```diff
@@ -9,6 +9,12 @@ export function expandMultipleValueVariable(input: string): string[] {
   return results;
 }
 
+const validTagTypes = ['DOUBLE', 'INT', 'U_INT64'];
+
+export function convertTagValue(type: string, value?: string) {
+  return value && validTagTypes.includes(type) ? Number(value) : value;
```
atmgrifter00, line 15: <nit> I think this reads better if you had something like:
```ts
const numericTagTypes = ['DOUBLE', 'INT', 'U_INT64'];

private function isNumericType(type: string): boolean {
    return numericTagTypes .includes(type);
}
    
export function convertTagValue(type: string, value?: string) {
  return value && isNumericType(type) ? Number(value) : value;
}
```

atmgrifter00 (commented):
Looks mostly good. Just waiting on comments to be resolved.

atmgrifter00 (approved):

src/datasources/tag/CurrentQueryHandler.ts
saradei-ni: Moved the logic for adding result fields in separate methods as suggested.

PR #155, opened 2025-04-29
fix(data-frame): Replace Variable support for $__all variable

richie-ni:
# Pull Request

## 🤨 Rationale
To fix : [Bug 3107967](https://dev.azure.com/ni/DevCentral/_workitems/edit/3107967): Dataframe data source doesn't handle Grafana variables when 'All' is selected as an option from the variable dropdown

We need to refactor the `replaceVariables` method to properly handle the `All` selection in the variables, as the `variable.current.value` of the `All` selection of variables is `$__all`, which is currently causing failures : 
image

## 👩‍💻 Implementation
- Refactored the logic of replacing the variables using `templateSrv.replace` instead of `template.Srv.getVariables` which does not list the proper value under `$__all` variable.
- Removed curly braces from multi-value variables as they return strings enclosed within curly braces

## 🧪 Testing

- Added unit test cases.
- Manually verified in Local env

## ✅ Checklist

- [ x ] This PR has a title that follows the [commit message format](https://github.com/ni/systemlink-grafana-plugins#commit-message-format).

### DIFF
--- a/src/core/utils.ts
+++ b/src/core/utils.ts
@@ -70,14 +70,13 @@ export function sleep(timeout: number) {
  */
 export function replaceVariables(values: string[], templateSrv: TemplateSrv) {
   const replaced: string[] = [];
-  values.forEach((col: string, index) => {
-    let value = col;
-    if (templateSrv.containsTemplate(col)) {
-      const variables = templateSrv.getVariables() as any[];
-      const variable = variables.find(v => v.name === col.split('$')[1]);
-      value = variable.current.value;
+  values.forEach(value => {
+    if (templateSrv.containsTemplate(value)) {
+      const variableReplacedValues = templateSrv.replace(value).replace(/[{}]/g, '').split(',');
+      replaced.push(...variableReplacedValues.filter(v => v.trim() !== ''));
+    } else {
+      replaced.push(value);
     }
-    replaced.push(value);
   });
   // Dedupe and flatten
   return [...new Set(replaced.flat())];
--- a/src/core/utils.ts
+++ b/src/core/utils.ts
@@ -72,7 +72,9 @@ export function replaceVariables(values: string[], templateSrv: TemplateSrv) {
   const replaced: string[] = [];
   values.forEach(value => {
     if (templateSrv.containsTemplate(value)) {
-      const variableReplacedValues = templateSrv.replace(value).replace(/[{}]/g, '').split(',');
+      const variableReplacedValues = templateSrv.replace(value) // Replace variable with their values
+        .replace(/[{}]/g, '') // return values without curly braces for multi-value variables which are returned as {value1,value2}
+        .split(',');
       replaced.push(...variableReplacedValues.filter(v => v.trim() !== ''));
     } else {
       replaced.push(value);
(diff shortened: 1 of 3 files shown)


ranjgith-ni (approved):
Add a little bit more context in the PR description on why we are moving away from the previous implementation of `variable.current.value`

atmgrifter00 (commented):
I think a comment in the code is all that is needed. I'll wait until I see the comment before approving.

src/core/utils.ts
```diff
@@ -70,14 +70,13 @@ export function sleep(timeout: number) {
...
   const replaced: string[] = [];
-  values.forEach((col: string, index) => {
-    let value = col;
-    if (templateSrv.containsTemplate(col)) {
-      const variables = templateSrv.getVariables() as any[];
-      const variable = variables.find(v => v.name === col.split('$')[1]);
-      value = variable.current.value;
+  values.forEach(value => {
+    if (templateSrv.containsTemplate(value)) {
+      const variableReplacedValues = templateSrv.replace(value).replace(/[{}]/g, '').split(',');
```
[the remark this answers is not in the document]
richie-ni: I have added a code comment with why the curl braces are removed from the variable.

atmgrifter00 (approved):

src/core/utils.ts
```diff
@@ -70,14 +70,15 @@ export function sleep(timeout: number) {
...
-  values.forEach((col: string, index) => {
-    let value = col;
-    if (templateSrv.containsTemplate(col)) {
-      const variables = templateSrv.getVariables() as any[];
-      const variable = variables.find(v => v.name === col.split('$')[1]);
-      value = variable.current.value;
+  values.forEach(value => {
+    if (templateSrv.containsTemplate(value)) {
+      const variableReplacedValues = templateSrv.replace(value) // Replace variable with their values
+        .replace(/[{}]/g, '') // return values without curly braces for multi-value variables which are returned as {value1,value2}
```
rajsite, line 76: It's not clear to me where the variable name syntax comes from so to ask at a high-level:

- Is it possible for the to be whitespace around the brackets that need to be accounted for, i.e. do these variable names come from user defined strings? `{ varname, varname }`
- Is it possible that the variable names themselves contain some of these characters like `{`, `}`, `,`

rajsite (approved):

src/core/utils.ts
richie-ni: @sudharsan-ni , 
Could you help clarify whether it’s a valid user scenario to have { or } within a variable name?

By default, when multi-value variables are selected, the returned value is enclosed in curly braces. For example, when using __$all, the values might appear as {value1, value2, value3, ...}. We currently strip the curly braces before creating the final list of variables.

sudharsan-ni: The values that we are passing here, are names of the columns in the data table. It is less likely for the column names to have these chars. 

But DFS API to create a data table doesn't do any validation to restrict these chars on column names. So as such, creating column names with these chars are not blocked. So, I'd suggest to handle it if it is a simple regex update or something of that sort. If not, we can skip.

richie-ni: A regex update wouldn’t solve the issue of variables wrapped in curly braces, since there’s no reliable way to distinguish between a multi-value variable and a simple variable that just happens to include curly braces.

richie-ni: Hence, we'll skip this case for now, and I’ll go ahead and close the PR.

PR #159, opened 2025-05-02
feat(results): Introduce feature flag toggles for Results and Steps query builders

Ahalya-ni:
# Pull Request

## 🤨 Rationale

As a part of the [User Story 2798322](https://ni.visualstudio.com/DevCentral/_workitems/edit/2798322): FE | Add Query Builder for Results Datasource
this PR introduces the ResultsConfigEditor to  toggle feature flags `queryByResults` and `queryBySteps`, enabling or disabling the visibility of the respective query builders.

## 👩‍💻 Implementation

- Added `ResultsConfigEditor` to having `DatasourceHttpSettings` and toggle switches to control the visibility of the Results and Steps Query Builders based on feature flags.
- Updated `module.ts` to set `ResultsConfigEditor` using `setConfigEditor` method and `ResultsDataSourceOptions` to define the datasource configuration type 
- Created required interfaces and constant for the feature flag toggles.
- Updated the Results Editor and Datasource to include the `ResultsDataSourceOptions` type, to ensure type safety.

## 🧪 Testing

- Added unit test

## ✅ Checklist

### DIFF
--- a/src/datasources/results/ResultsConfigEditor.test.tsx
+++ b/src/datasources/results/ResultsConfigEditor.test.tsx
@@ -33,16 +33,16 @@ describe('ResultsConfigEditor', () => {
     stepsQueryBuilder = screen.getAllByRole('checkbox')[1];
     jest.clearAllMocks();
   });
-  test('renders DataSourceHttpSettings component', () => {
+  test('should renders DataSourceHttpSettings component when ResultsConfigEditor is loaded', () => {
     expect(screen.getByText('Mock DataSourceHttpSettings')).toBeInTheDocument();
   });
 
-  test('renders the component with feature toggles', () => {
+  test('should render the component with feature toggles when loaded', () => {
     expect(resultsQueryBuilder).toBeInTheDocument();
     expect(stepsQueryBuilder).toBeInTheDocument();
   });
 
-  test('toggles Results Query Builder feature', async () => {
+  test('Should update the queryByResults feature toggle when it is toggled', async () => {
     expect(resultsQueryBuilder).not.toBeChecked();
 
     await userEvent.click(resultsQueryBuilder);
@@ -53,7 +53,7 @@ describe('ResultsConfigEditor', () => {
     });
   });
 
-  test('toggles Steps Query Builder feature', () => {
+  test('Should update the queryBySteps feature toggle when it is toggled', () => {
     expect(stepsQueryBuilder).not.toBeChecked();
 
     fireEvent.click(stepsQueryBuilder);
--- a/src/datasources/results/ResultsConfigEditor.test.tsx
+++ b/src/datasources/results/ResultsConfigEditor.test.tsx
@@ -1,10 +1,14 @@
 import React from 'react';
 import { render, screen, fireEvent, waitFor } from '@testing-library/react';
 import { ResultsConfigEditor } from './ResultsConfigEditor';
-import { DataSourcePluginOptionsEditorProps } from '@grafana/data';
+import { DataSourcePluginOptionsEditorProps, DataSourceSettings } from '@grafana/data';
 import userEvent from '@testing-library/user-event';
 
 const mockOnOptionsChange = jest.fn();
+jest.mock('@grafana/ui', () => ({
+  ...jest.requireActual('@grafana/ui'),
+  DataSourceHttpSettings: jest.fn(() => <div>Mock DataSourceHttpSettings</div>),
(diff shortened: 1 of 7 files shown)


src/datasources/results/ResultsConfigEditor.tsx
```diff
@@ -0,0 +1,53 @@
...
+          <Text element="h6">Features</Text>
+        </div>
+        <InlineSegmentGroup>
+          <InlineField label="Results Query Builder" labelWidth={25}>
+            <InlineSwitch
+              value={options.jsonData?.featureToggles?.queryByResults ?? ResultsFeatureTogglesDefaults.queryByResults}
+              onChange={handleFeatureChange('queryByResults')}
+            />
+          </InlineField>
+          <Tag name="Beta" colorIndex={5} />
```
richie-ni, line 39: is it still necessary to use Beta...since Assets were the first to implement FF ..they had it put as Beta

src/datasources/results/ResultsConfigEditor.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+describe('ResultsConfigEditor', () => {
+  beforeEach(() => {
+    render(<ResultsConfigEditor {...defaultProps} />);
+
+    resultsQueryBuilder = screen.getAllByRole('checkbox')[0];
+    stepsQueryBuilder = screen.getAllByRole('checkbox')[1];
+    jest.clearAllMocks();
+  });
+  test('renders DataSourceHttpSettings component', () => {
```
richie-ni, line 36: Lets use the `Should-When-Then` format for the test title

richie-ni (approved):

src/datasources/results/ResultsConfigEditor.test.tsx
Ahalya-ni: Updated

src/datasources/results/ResultsConfigEditor.tsx
Ahalya-ni: No. I have removed Beta

src/datasources/results/ResultsConfigEditor.tsx
```diff
@@ -0,0 +1,51 @@
+import React, { ChangeEvent, useCallback } from 'react';
+import { DataSourcePluginOptionsEditorProps } from '@grafana/data';
+import { DataSourceHttpSettings, InlineField, InlineSegmentGroup, InlineSwitch, Text } from '@grafana/ui';
+import { ResultsDataSourceOptions, ResultsFeatureTogglesDefaults } from './types/types';
+
+interface Props extends DataSourcePluginOptionsEditorProps<ResultsDataSourceOptions> {}
+
+export const ResultsConfigEditor: React.FC<Props> = ({ options, onOptionsChange }) => {
+    const handleFeatureChange = useCallback(
```
ranjgith-ni, line 9: Why do we need a [`useCallback()`](https://react.dev/reference/react/useCallback) hook here? 

We are binding this method in an `onChange` event - which should not re-render the page, and I do not see any other element that would get refreshed by the changes done in this method

src/datasources/results/ResultsConfigEditor.tsx
```diff
@@ -0,0 +1,51 @@
...
+    (featureKey: string) => (event: ChangeEvent<HTMLInputElement>) => {
+      const jsonData = {
+        ...options.jsonData,
+        ...{ featureToggles: { ...options.jsonData.featureToggles, [featureKey]: event.target.checked } },
```
ranjgith-ni, line 13:
```suggestion
        featureToggles: { ...options.jsonData.featureToggles, [featureKey]: event.target.checked },
```

Why do we need a copy here?

ranjgith-ni (approved):

src/datasources/results/ResultsConfigEditor.tsx
Ahalya-ni: Yes that i sright, since `handleFeatureChange` is not passed as child prop to components or used in dependency array, `useCallback` is not required in this case.

src/datasources/results/ResultsConfigEditor.tsx
Ahalya-ni: The spread operator is not necessary because we are already creating a new featureToggles object with `{ ...options.jsonData.featureToggles, [featureKey]: event.target.checked }`

PR #161, opened 2025-05-02
feat: Add sharable cache

priya-ni:
# Pull Request

## 🤨 Rationale

To add a new singleton class that will be used as sharable cache in upcoming datasources

## 👩‍💻 Implementation
- 

## 🧪 Testing

Added unit tests

## ✅ Checklist

### DIFF
--- a/src/core/shared-cache.ts
+++ b/src/core/shared-cache.ts
@@ -1,0 +1,46 @@
+/**
+ * A singleton class for managing a shared in-memory cache.
+ * This class ensures that only one instance of the cache exists across the application.
+ */
+class SharedCache {
+  private static instance: SharedCache; // Holds the singleton instance
+  private store: Map<string, any> = new Map(); // Internal storage for cached data
+
+  /**
+   * Retrieves the singleton instance of the SharedCache.
+   * If the instance does not exist, it creates one.
+   * 
+   * @returns {SharedCache} The singleton instance of the SharedCache.
+   */
+  public static getInstance (): SharedCache {
+    if (!SharedCache.instance) {
+      SharedCache.instance = new SharedCache();
+    }
+    return SharedCache.instance;
+  }
+
+  /**
+   * Retrieves a value from the cache by its key.
+   * 
+   * @param {string} key - The key of the cached value.
+   * @returns {any} The cached value, or undefined if the key does not exist.
+   */
+  get (key: string): any {
+    return this.store.get(key);
+  }
+
+  /**
+   * Stores a value in the cache with the specified key.
+   * 
+   * @param {string} key - The key to associate with the cached value.
+   * @param {any} data - The value to store in the cache.
+   */
+  set (key: string, data: any): void {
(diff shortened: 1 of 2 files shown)


manisha-ni (approved):

src/core/shared-cache.ts
```diff
@@ -0,0 +1,46 @@
+/**
+ * A singleton class for managing a shared in-memory cache.
+ * This class ensures that only one instance of the cache exists across the application.
+ */
+class SharedCache {
```
atmgrifter00, line 5: So what is the driving force behind needing this? Is there a spec somewhere that describes the need? Naively, something like this seems a bit extreme, in that it seems to be tackling a problem in as broad and heavy-handed a way as possible. Things to consider:

- Any new public API that allows `any` as part of its contract is questionable.
- Providing a cache mechanism with no automatic means of clean-up should be thought through and ultimately explicitly stated why it is unavoidable.

atmgrifter00 (commented):
Have questions regarding this change that need to be answered.

src/core/shared-cache.ts
rajsite: +1 a giant shared mutable global variable holding untyped state sounds very difficult to maintain. Would also like to understand the context like meyers described

src/core/shared-cache.ts
```diff
@@ -0,0 +1,103 @@
...
+    this.store.set(key, { value: data, expiresAt });
+    this.startCleanup();
+  }
+
+  /**
+   * Removes expired keys from the cache.
+   */
+  private cleanup(): void {
+    const now = Date.now();
+    for (const [key, { expiresAt }] of this.store.entries()) {
```
ranjgith-ni, line 68: nit: use `forEach`

ranjgith-ni (approved):

src/core/shared-cache.ts
priya-ni: I have updated the description to include the HLD link & rationale.

In addition, for clearing the value automatically `TimeToLive` is introduced in the cache.

priya-ni:
@rajsite @atmgrifter00 @cameronwaterman Could you please review this PR?

atmgrifter00:
@priya-ni, I think there are still some very fundamental issues with the approach you are taking here. First, the fact that the `SharedCache` is generic doesn't really provide any significant difference in its typing, and in fact, is still used effectively with `any` for its value typing. The cleanup code running on a timer also is concerning, as this seems like it could be the source of unexpected behaviors/bugs.

Ultimately, it would be good to understand if there are patterns/mechanisms that the frameworks we are building on top of (i.e. Grafana/React) provide for us to handle concerns such as this. It seems like what we want is a dependency injection system with a service that can then be injected where it is needed (as we do in Angular).

I believe we intend to discuss this PR at tomorrow's Soliton Tech Syncup (05/14). Hopefully after that meeting we can provide some better direction.

PR #175, opened 2025-05-09
feat: add nimble theming for query builder

manisha-ni:
# Pull Request

## 🤨 Rationale
To add stylings for query builder so that it more in balance with other elements of datasource

## 👩‍💻 Implementation

- Added custom css file with nimble theme for primary color and other styles in sync with other datasource fields

theming of datasource2

**Light Theme**
image
image

**Dark Theme**
image
image

## 🧪 Testing

Manually verified

## ✅ Checklist

### DIFF
--- a/src/core/components/SlQueryBuilder/SlQueryBuilder.css
+++ b/src/core/components/SlQueryBuilder/SlQueryBuilder.css
@@ -6,7 +6,6 @@ body {
     --smart-border-bottom-right-radius: 2px;
     --smart-border-bottom-left-radius: 2px;
     --smart-ui-state-color-focus: var(--smart-background-color);
-    /* --smart-font-family: Inter, Helvetica, Arial, sans-serif; */
 }
 
 body[theme="sl-light"], .smart-ui-component[theme="sl-light"] {
--- a/src/core/components/SlQueryBuilder/SlQueryBuilder.css
+++ b/src/core/components/SlQueryBuilder/SlQueryBuilder.css
@@ -1,45 +1,33 @@
+@import url('smart-webcomponents-react/source/styles/components/smart.base.css');
+@import url('smart-webcomponents-react/source/styles/components/smart.common.css');
+@import url('smart-webcomponents-react/source/styles/components/smart.querybuilder.css');
 @import url('@ni/nimble-tokens/dist/styledictionary/css/variables.css');
 
-body {
+.smart-query-builder {
     --smart-primary: var(--ni-nimble-base-digital-green-light);
-    --smart-disabled: #4D5359;
-    --smart-border-top-right-radius: 2px;
-    --smart-border-top-left-radius: 2px;
-    --smart-border-bottom-right-radius: 2px;
-    --smart-border-bottom-left-radius: 2px;
+
+    --smart-background: var(--grafana-background-primary);
     --smart-ui-state-color-focus: var(--smart-background-color);
+    --smart-border: var(--grafana-border-medium);
+    --smart-background-color: var(--grafana-text-primary);
+    --smart-ui-state-focus: var(--grafana-focus-color);
+    --smart-border-top-right-radius: var(--grafana-border-radius-default);
+    --smart-border-top-left-radius: var(--grafana-border-radius-default);
+    --smart-border-bottom-right-radius: var(--grafana-border-radius-default);
+    --smart-border-bottom-left-radius: var(--grafana-border-radius-default);
 }
 
-body[theme="sl-light"], .smart-ui-component[theme="sl-light"] {
-    --smart-background: #ffffff;
-    --smart-background-color: #24292e;
-    --smart-ui-state-focus: #1616171a;
-    --smart-border: #24292e4d;
-}
-
-body[theme="sl-dark"], .smart-ui-component[theme="sl-dark"] {
-    --smart-background:#111217;
-    --smart-background-color: #ccccdc;
-    --smart-primary-color:#ffffff;
-    --smart-border: #ccccdc33;
-    --smart-ui-state-focus: #f1f1f21a;
-
-    --smart-scroll-bar-background: #3E3E42;
-    --smart-scroll-bar-track-background: #3E3E42;
-    --smart-scroll-bar-border: #3E3E42;
-    --smart-scroll-bar-thumb-background: #686868;
-    --smart-scroll-bar-thumb-border: #686868;
-    --smart-scroll-bar-thumb-background-hover: #9E9E9E;
-    --smart-scroll-bar-thumb-border-hover: #9E9E9E;
-    --smart-scroll-bar-thumb-background-active: #444;
-    --smart-scroll-bar-thumb-border-active: #444;
-    --smart-scroll-bar-button-background: #3E3E42;
-    --smart-scroll-bar-button-border: #3E3E42;
-    --smart-scroll-bar-button-color: #999999;
-    --smart-scroll-bar-button-background-hover: #3E3E42;
-    --smart-scroll-bar-button-border-hover: #3E3E42;
-    --smart-scroll-bar-button-color-hover: var(--smart-primary);
-    --smart-scroll-bar-button-background-active: #3E3E42;
-    --smart-scroll-bar-button-border-active: #3E3E42;
-    --smart-scroll-bar-button-color-active: var(--smart-primary);
+.smart-scroll-bar {
+    --smart-scroll-bar-button-border: transparent;
+    --smart-scroll-bar-background: var(--smart-background);
+    --smart-scroll-bar-track-background: var(--grafana-focus-color);
+    --smart-scroll-bar-thumb-background: var(--grafana-focus-color);
+    --smart-scroll-bar-thumb-background-hover: var(--grafana-focus-color);
+    --smart-scroll-bar-thumb-background-active: var(--grafana-active-color);
+    --smart-scroll-bar-button-background: var(--grafana-focus-color);
+    --smart-scroll-bar-button-color: var(--grafana-text-primary);
+    --smart-scroll-bar-button-background-hover: var(--grafana-focus-color);
+    --smart-scroll-bar-button-color-hover: var(--grafana-text-primary);
+    --smart-scroll-bar-button-background-active: var(--grafana-focus-color);
+    --smart-scroll-bar-button-color-active: var(--grafana-text-primary);
 }
--- a/src/core/components/SlQueryBuilder/SlQueryBuilder.css
+++ b/src/core/components/SlQueryBuilder/SlQueryBuilder.css
@@ -1,5 +1,7 @@
+@import url('@ni/nimble-tokens/dist/styledictionary/css/variables.css');
+
 body {
-    --smart-primary: #009b65;
+    --smart-primary: var(--ni-nimble-base-digital-green-light);
     --smart-disabled: #4D5359;
     --smart-border-top-right-radius: 2px;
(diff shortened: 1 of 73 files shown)


ranjgith-ni (approved):

manisha-ni:
@rajsite @atmgrifter00 @cameronwaterman

Please review this PR.

src/core/components/SlQueryBuilder/SlQueryBuilder.css
```diff
@@ -0,0 +1,43 @@
+body {
+    --smart-primary: #009b65;
```
atmgrifter00, line 2: Can't we bring in a dependency to @ni/nimble-tokens, and then actually use the token values here?

manisha-ni: @atmgrifter00 - Yes, but I was only able to use the base tokens.

Anyways, we use only the nimble color for the background of the add icon. Other stylings are more inline with grafana's theme as having nimble theming looked awkward with other controls

atmgrifter00: Looks fine to me. I'll defer to @rajsite for any other needed changes.

atmgrifter00 (approved):

src/core/components/SlQueryBuilder/SlQueryBuilder.tsx
```diff
@@ -1,11 +1,10 @@
 import React, { useMemo } from 'react';
 import { useTheme2 } from '@grafana/ui';
 import QueryBuilder, { QueryBuilderProps } from 'smart-webcomponents-react/querybuilder';
-import 'smart-webcomponents-react/source/styles/smart.dark-orange.css';
-import 'smart-webcomponents-react/source/styles/smart.orange.css';
 import 'smart-webcomponents-react/source/styles/components/smart.base.css';
 import 'smart-webcomponents-react/source/styles/components/smart.common.css';
 import 'smart-webcomponents-react/source/styles/components/smart.querybuilder.css';
```
rajsite, line 6: Can these be moved into SLQueryBuilder.css as well?

src/core/components/SlQueryBuilder/SlQueryBuilder.css
```diff
@@ -0,0 +1,45 @@
+@import url('@ni/nimble-tokens/dist/styledictionary/css/variables.css');
+
+body {
```
rajsite, line 3: There are a lot of duplicated constants and no comments about where thee values came from.

For the colors in particular can you make named aliases and comments to where these values are coming from?

```css
body {
   /* Tokens from: <src_url> */
    --ni-grafana-query-builder-blue: #3E3E42
}

/* ... */ {
    --smart-scroll-bar-background: var(--ni-grafana-query-builder-blue);
    --smart-scroll-bar-track-background:  var(--ni-grafana-query-builder-blue);
}
```

src/core/components/SlQueryBuilder/SlQueryBuilder.tsx
```diff
@@ -36,7 +35,7 @@ export const SlQueryBuilder: React.FC<SlQueryBuilderProps> = ({
   validateOnInput = false,
 }) => {
   const theme = useTheme2();
-  document.body.setAttribute('theme', theme.isDark ? 'dark-orange' : 'orange');
+  document.body.setAttribute('theme', theme.isDark ? 'sl-dark' : 'sl-light');
```
rajsite, line 38: Instead of creating global styles and mutating elements we don't own (i.e. the body element), can we instead set CSS Custom Properties on the QueryBuilder element itself?

I don't know the exact React syntax but conceptually something like:
```tsx
<QueryBuilder
    style={{
      --smart-background: <value>
   }}
```

Can then dedupe properties:
```tsx
const myBlue;
<QueryBuilder
    style={{
        '--smart-scroll-bar-background': myBlue,
        '--smart-scroll-bar-track-background': myBlue,
   }}
```

Also if the current hard coded / custom properties are from the garfana theme, it looks like we can plugin to the theme system: https://github.com/grafana/grafana/blob/main/contribute/style-guides/themes.md

```tsx
const theme = useTheme2();
<QueryBuilder
    style={{
        '--smart-scroll-bar-background': theme.colors.background.primary
   }}
```

manisha-ni: @rajsite I tried what you suggested but assigning values to the variables did not work as expected.

I've used `style.setProperty` to set the values of grafana theme to a variable and used that variable to apply theme instead. 

Please take a look and let me know your thoughts

src/core/components/SlQueryBuilder/SlQueryBuilder.tsx
```diff
@@ -36,7 +32,15 @@ export const SlQueryBuilder: React.FC<SlQueryBuilderProps> = ({
   validateOnInput = false,
 }) => {
   const theme = useTheme2();
-  document.body.setAttribute('theme', theme.isDark ? 'dark-orange' : 'orange');
+
+  useEffect(() => {
+    document.body.style.setProperty('--grafana-background-primary', theme.colors.background.primary);
```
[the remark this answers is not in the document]
manisha-ni: @atmgrifter00 Having these properties in inline style did not work.

I tried to using 'style.setProperty` with query builder reference to set these properties but query builder reference doesn't have style. So I have set the property in body style and used class names in css for styling

src/core/components/SlQueryBuilder/SlQueryBuilder.tsx
rajsite: > I tried what you suggested but assigning values to the variables did not work as expected.

My feedback was:

> I don't know the exact React syntax but conceptually something like:

Can you explain more what you tried / researched for React? As mentioned I don't know what the expected syntax is.

> the variables did not work as expected.the variables did not work as expected.

What didn't work exactly? This isn't a very clear statement

src/core/components/SlQueryBuilder/SlQueryBuilder.css
```diff
@@ -0,0 +1,38 @@
...
+@import url('smart-webcomponents-react/source/styles/components/smart.common.css');
+@import url('smart-webcomponents-react/source/styles/components/smart.querybuilder.css');
+@import url('@ni/nimble-tokens/dist/styledictionary/css/variables.css');
+
+.smart-query-builder {
+    /* Tokens from nimble tokens */
+    --smart-primary: var(--ni-nimble-base-digital-green-light);
+
+    /* Tokens from grafana */
+    --smart-background: var(--grafana-background-primary);
```
rajsite, line 11: Use the prefix `--ni-grafana` to emphasize that these are custom properties invented and defined by NI and are not defined by grafana.

src/core/components/SlQueryBuilder/SlQueryBuilder.tsx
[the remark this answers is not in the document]
rajsite: You didn't describe what didn't work so it's hard to provide a useful suggestion. A random one can be to wrap the query bulder in a `div` and set the custom property value style on that:
```
<div style={{/**/}}>
   <QueryBuilder>...
</div>
```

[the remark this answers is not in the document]
atmgrifter00: Here's a [SO post](https://stackoverflow.com/questions/52005083/how-to-define-css-variables-in-style-attribute-in-react-and-typescript) that reveals the magic syntax.

src/core/components/SlQueryBuilder/SlQueryBuilder.tsx
manisha-ni: Sorry, I was vague about what did not work. 

But basically the 'style' in the QueryBuilder node did not compute at all. So whatever variable we defined there was not computed and so the style did not apply. 
I'm not sure exactly why this is so but I think it may be because QueryBuilder is not exactly an HTML element

I also tried getting the refence from query builder using [useRef](https://react.dev/reference/react/useRef) to apply styles to it. But again, it doesn't have style property so I was unsuccessful

src/core/components/SlQueryBuilder/SlQueryBuilder.tsx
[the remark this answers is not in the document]
manisha-ni: Thanks for the suggestion @rajsite, having div around the query builder and computing variables in it's style property worked!

src/core/components/SlQueryBuilder/SlQueryBuilder.css
manisha-ni: Updated the prefix

src/core/components/SlQueryBuilder/SlQueryBuilder.css
```diff
@@ -0,0 +1,38 @@
...
+    --smart-ui-state-focus: var(--ni-grafana-focus-color);
+    --smart-border-top-right-radius: var(--ni-grafana-border-radius-default);
+    --smart-border-top-left-radius: var(--ni-grafana-border-radius-default);
+    --smart-border-bottom-right-radius: var(--ni-grafana-border-radius-default);
+    --smart-border-bottom-left-radius: var(--ni-grafana-border-radius-default);
+
+    --smart-ui-state-color-focus: var(--smart-background-color);
+}
+
+.sl-query-builder, .smart-scroll-bar {
```
rajsite, line 23: Are the differences between these selectors intentional?

`.sl-query-builder .smart-query-builder` i.e. `.smart-query-builder` is a child of `.sl-query-builder` in the page

vs

Apply all these styles to both `.sl-query-builder` and `.smart-scroll-bar` separately.

Seems unexpected, is the scrollbar not a child of the query builder? Should this be:

```suggestion
.sl-query-builder .smart-scroll-bar {
```

rajsite (approved):

src/core/components/SlQueryBuilder/SlQueryBuilder.tsx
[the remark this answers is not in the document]
manisha-ni: @rajsite @atmgrifter00 
While with `div` the issue of adding styles to the query builder was solved, the defined variables were not accessible by the scroll viewer (which loads the options). This is because the scroll viewer is not a child of query builder but of body
image
So, we need to define the variables in `body` level so that the scroll viewer can access it too

Also, the styles we import are being set globally. So, if other query builders, not using sl-query-builder, are opened first, the theme set in body will apply to the sl-query-builder and only the properties we set is overwritten. This was the reason I didn't see the issue with scroll viewer when I was testing

Similarly, we can't apply the style for scroll viewer to a class defined by us and hence we should use the common `.smart-scroll-viewer` class. So, the styles we apply are being carried over to other query builders leading to unwanted change as the variables will not be defined for them

To solve these issues, I have kept the theme attribute so no styles are spilled over.

I am closing this PR now. If you have any concerns do let me know, I will take it in subsequent PR

cc - @ranjgith-ni @priya-ni

src/core/components/SlQueryBuilder/SlQueryBuilder.tsx
```diff
@@ -36,7 +33,26 @@ export const SlQueryBuilder: React.FC<SlQueryBuilderProps> = ({
   validateOnInput = false,
 }) => {
   const theme = useTheme2();
-  document.body.setAttribute('theme', theme.isDark ? 'dark-orange' : 'orange');
+
+  useEffect(() => {
+    document.body.setAttribute('theme', 'ni-grafana');
```
rajsite, line 38: What is the point of this?

PR #179, opened 2025-05-12
fix(results): Update keyword filter operations in result query builder

Ahalya-ni:
# Pull Request

## 🤨 Rationale

To address the issue with the Keyword field filter operations, as it represents a list rather than a string, the filter operations have been updated to handle list-based operations instead of string-based ones.
image

## 👩‍💻 Implementation

- Updated the filter operations for the Keyword field to handle it as a list instead of a string.
- Added the respective list-based operations to the custom operations array in the query builder.

## 🧪 Testing

Yet to add

## ✅ Checklist

### DIFF
--- a/src/datasources/results/components/query-builders/query-results/ResultsQueryBuilder.test.tsx
+++ b/src/datasources/results/components/query-builders/query-results/ResultsQueryBuilder.test.tsx
@@ -56,10 +56,10 @@ describe('ResultsQueryBuilder', () => {
     });
 
     it('should select keyword in query builder', () => {
-      const { conditionsContainer } = renderElement([], [], [], 'Keywords listequals "keyword1"');
+      const { conditionsContainer } = renderElement([], [], [], 'Keywords.Contains("keyword1")');
 
       expect(conditionsContainer?.length).toBe(1);
-      expect(conditionsContainer.item(0)?.textContent).toContain("Keywords"); //label
+      expect(conditionsContainer.item(0)?.textContent).toContain("Keyword"); //label
       expect(conditionsContainer.item(0)?.textContent).toContain("Equals"); //operator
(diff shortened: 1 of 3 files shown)


src/datasources/results/components/query-builders/query-results/ResultsQueryBuilder.test.tsx
```diff
@@ -55,6 +55,15 @@ describe('ResultsQueryBuilder', () => {
       expect(conditionsContainer.item(0)?.textContent).toContain("PASSED"); //value
     });
 
+    it('should select keyword in query builder', () => {
+      const { conditionsContainer } = renderElement([], [], [], 'Keywords.Contains("keyword1")');
+
+      expect(conditionsContainer?.length).toBe(1);
+      expect(conditionsContainer.item(0)?.textContent).toContain("Keyword"); //label
```
richie-ni, line 62:
```suggestion
      expect(conditionsContainer.item(0)?.textContent).toContain("Keywords"); //label
```
Why is the name in the property dropdown `Keyword` and not `Keywords`?

richie-ni (approved):

src/datasources/results/components/query-builders/query-results/ResultsQueryBuilder.test.tsx
Ahalya-ni: In SLE Results query builder, the label is `Keyword`
image

But in results query builder of dataspace query it is `Keywords`. Should I provide labels similar to dataspace or results query builder?

ranjgith-ni (approved):

src/datasources/results/components/query-builders/query-results/ResultsQueryBuilder.test.tsx
ranjgith-ni: I am okay with keeping it as Keyword

PR #195, opened 2025-05-20
fix(results): Label not persisting in Query Builder after selection from dropdown

Ahalya-ni:
# Pull Request

## 🤨 Rationale
As a part of this [User Story 2798322](https://ni.visualstudio.com/DevCentral/_workitems/edit/2798322): FE | Add Query Builder for Results Datasource,

this PR addresses the issue where the label in the query builder was not persisting after being selected from the dropdown.

## 👩‍💻 Implementation
- Updated `QueryResultsEditor`and `StepsQueryBuilderWrapper` to introduce a state variable called `areDependenciesLoaded` which is initially set to false.
- Once all dependencies (e.g., workspaces, part numbers) are loaded, the `areDependenciesLoaded` state os updated to true and passed to the respective query builders.
- Modified `ResultsQueryBuilder` and `StepsQueryBuilder` to conditionally load fields and operators based on the areDependenciesLoaded state. This ensures that fields and operators are only displayed after all dependencies are fully loaded.

Note:
In `ResultsVariableQueryEditor`, as the component is still under development and there are no dependencies to load at the moment, the `areDependenciesLoaded` field, which is required, has been set to true when the component is mounted. This is a temporary implementation to handle the required field, and it will be updated in the future when actual dependencies are introduced.

## 🧪 Testing

- Added unit test

## ✅ Checklist

### DIFF
--- a/src/datasources/results/ResultsDataSourceBase.ts
+++ b/src/datasources/results/ResultsDataSourceBase.ts
@@ -7,8 +7,15 @@ import { ResultsPropertiesOptions } from "./types/QueryResults.types";
 import { getVariableOptions } from "core/utils";
 import { ExpressionTransformFunction } from "core/query-builder.utils";
 import { QueryBuilderOperations } from "core/query-builder.constants";
+import { parseErrorMessage } from "core/errors";
 
 export abstract class ResultsDataSourceBase extends DataSourceBase<ResultsQuery> {
+  private partNumbersLoaded!: () => void;
+  private workspacesLoaded!: () => void;
+
+  public arePartNumbersLoaded$ = new Promise<void>(resolve => this.partNumbersLoaded = resolve);
+  public areWorkspacesLoaded$ = new Promise<void>(resolve => this.workspacesLoaded = resolve);
+
   baseUrl = this.instanceSettings.url + '/nitestmonitor';
   queryResultsValuesUrl = this.baseUrl + '/v2/query-result-values';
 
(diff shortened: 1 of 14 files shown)


src/datasources/results/ResultsDataSourceBase.ts
```diff
@@ -118,10 +132,12 @@ export abstract class ResultsDataSourceBase extends DataSourceBase<ResultsQuery>
     const partNumbers = await this.post<string[]>(this.queryResultsValuesUrl, {
       field: ResultsPropertiesOptions.PART_NUMBER,
     }).catch(error => {
-      throw new Error(error);
+      console.error('Error fetching part numbers:', parseErrorMessage(error));
```
Ahalya-ni, line 135: This change allows subsequent API calls to continue even if one of the API calls fails. Instead of throwing an error, the failure is logged using `console.error`, preventing the execution flow from being interrupted. In the future, `console.error` will be replaced with a more comprehensive error-handling mechanism.

src/datasources/results/ResultsDataSourceBase.ts
```diff
@@ -104,7 +104,7 @@ export abstract class ResultsDataSourceBase extends DataSourceBase<ResultsQuery>
 
     const workspaces = await this.getWorkspaces()
       .catch(error => {
-        throw new Error(error);
+        return;
```
Ahalya-ni, line 107: This change allows subsequent API calls to continue even if one of the API calls fails. Instead of throwing an error, the failure is returned, preventing the execution flow from being interrupted. In the future, it will be replaced with a more comprehensive error-handling mechanism.

src/datasources/results/ResultsDataSourceBase.ts
```diff
@@ -118,7 +118,7 @@ export abstract class ResultsDataSourceBase extends DataSourceBase<ResultsQuery>
     const partNumbers = await this.post<string[]>(this.queryResultsValuesUrl, {
       field: ResultsPropertiesOptions.PART_NUMBER,
     }).catch(error => {
-      throw new Error(error);
+      console.error('Error in loading part numbers:', error);
```
Ahalya-ni, line 121: This change—from `throw new Error` to `console.error` is necessary to allow the flow to continue even if one of the dependencies (workspace or part number) fails. The current use of `console.error` is a temporary measure; it will be replaced with a proper error handling mechanism in the future, and a [user story](https://ni.visualstudio.com/DevCentral/_workitems/edit/3107677) has been created to track this implementation.

richie-ni (approved):
Could we add up in the implementation the reason why we add null and how null helps us to solve the issue

PR #199, opened 2025-05-21
feat(test-plans): Query test plans in batches

priya-ni:
# Pull Request

## 🤨 Rationale
[Task 3106073](https://ni.visualstudio.com/DevCentral/_workitems/edit/3106073): Add support to query testplans
To query the test plans in batches.

## 👩‍💻 Implementation
- Moved `queryInBatches` from `ResultsDataSourceBase` to `core/utils`
- Updated `runQuery` of `TestPlansDataSource` to query the values in batches
- Updated properties & projection maps for querying

## 🧪 Testing
- Added unit tests

## ✅ Checklist


src/datasources/test-plans/types.ts
```diff
@@ -14,61 +14,166 @@ export enum OutputType {
...
-    FIXTURE_IDS: 'fixtureIds',
-    DUT: 'dut',
+export enum TimeFields {
+    CREATED_AT = 'CREATED_AT',
+    UPDATED_AT = 'UPDATED_AT',
+    PLANNED_START_DATE_TIME = 'PLANNED_START_DATE_TIME',
+    ESTIMATED_END_DATE_TIME = 'ESTIMATED_END_DATE_TIME'
+};
+
+export const PropertiesProjectionMap: Record<Properties, { label: string, projection: string[], field: string[] }> = {
```
rajsite, line 51: We should improve the typing of this object instead of just being "string".

I'm making a guess, but looks like the "projection" is associated with the Properties enum and the "field" is TestPlanResponseProperties. If that is the case we can get typescript to represent that as:

```suggestion
export const PropertiesProjectionMap: Record<Properties, {
    label: string,
    projection: (keyof typeof Properties)[],
    field: (keyof TestPlanResponseProperties)[]
}> = {
```

And if that is accurate then it looks like the object has a few errors:

image

image

src/datasources/test-plans/types.ts
```diff
@@ -14,61 +14,166 @@ export enum OutputType {
...
+        label: 'Fixture names',
+        projection: ['FIXTURE_IDS'],
+        field: ['fixtureIds'],
+    },
+    [Properties.DUT_ID]: {
+        label: 'DUT',
+        projection: ['DUT_ID'],
+        field: ['dutId'],
+    },
 };
```
rajsite, line 177: Should mark exported const objects `as cont`

```suggestion
} as const;
```

src/datasources/test-plans/types.ts
```diff
@@ -88,3 +193,38 @@ export const OrderByOptions = {
     ID: 'ID',
     UPDATED_AT: 'UPDATED_AT'
 };
```
rajsite, line 195: Should mark exported const objects `as cont`

```suggestion
} as const;
```

src/datasources/test-plans/types.ts
```diff
@@ -14,61 +14,166 @@ export enum OutputType {
...
+        field: ['fixtureIds'],
+    },
+    [Properties.DUT_ID]: {
+        label: 'DUT',
+        projection: ['DUT_ID'],
+        field: ['dutId'],
+    },
 };
 
 export const OrderBy = [
```
rajsite, line 179: I can't mark the line but this object should also be `as const`

src/datasources/test-plans/types.ts
priya-ni: The assumption is partially correct that `projection` is not `(keyof typeof Properties)`, because the type `Properties` used to show UI dropdown, but the actual API does not allow those values as projection. I will create a separate enum for `Projection` and use it here.

src/datasources/test-plans/types.ts
priya-ni: Applied suggestion.

src/datasources/test-plans/types.ts
priya-ni: Applied suggestion.

priya-ni:
@rajsite @atmgrifter00 @cameronwaterman Could you please review this PR?

PR #213, opened 2025-05-23
fix(results): dependency load 

Ahalya-ni:
# Pull Request

## 🤨 Rationale

This PR ensures that dependency calls such as `query-result-values` (for part numbers) and `user` (for workspaces)  are made only once when test results datasource is loaded. This ensures caching of dependency data across different query types `results` and `steps`.

## 👩‍💻 Implementation

- Introduced static variables `workspacePromise` and `partNumberPromise` to guarantee that each dependency API call is made only once and shared across all instances.
- Made the cache variables for workspaces and part numbers static, so cached data is shared globally within the plugin.
- Updated `loadWorkspaces` and `getPartNumbers` to check 
    - Check if the corresponding promise already exists and return it if so.
    - If not, initiate the API call, store the resulting promise, and update the static cache.
- Updated editors to use static cache values
- Modified data source constructors to trigger the loading of these dependencies at instantiation.

## 🧪 Testing

Added unit test

## ✅ Checklist


src/datasources/results/ResultsDataSourceBase.ts
```diff
@@ -21,8 +21,9 @@ export abstract class ResultsDataSourceBase extends DataSourceBase<ResultsQuery>
   private toDateString = '${__to:date}';
 
   readonly globalVariableOptions = (): QueryBuilderOption[] => getVariableOptions(this);
-  readonly workspacesCache = new Map<string, Workspace>([]);
-  readonly partNumbersCache: string[] = [];
+
+  static workspacesPromise: Promise<Map<string, Workspace> | void> | null = null;
+  static partNumbersPromise: Promise<string[] | void> | null = null;
```
ranjgith-ni, line 26: await: Let's not continue with the `null` alternate for the promise, instead assign the promise directly to the `readonly` property


```suggestion
  readonly workspacesCache: Promise<Map<string, Workspace>> = this.getWorkspaces()
      .then(workspaces => {
        const workspaceMap = new Map<string, Workspace>();
        if (workspaces) {
          workspaces.forEach(workspace => workspaceMap.set(workspace.id, workspace));
        }
        return workspaceMap;
      })
      .catch(error => {
        console.error('Error in loading workspaces:', error);
        return new Map<string, Workspace>();
      });

  readonly partNumbersCache: Promise<string[]> = this.queryResultsValues(ResultsPropertiesOptions.PART_NUMBER, undefined)
      .catch(error => {
        console.error('Error in loading part numbers:', error);
        return [];
      });
```

This would make sure to keep the consumption of these caches clean

Ahalya-ni: I understand your suggestion to initialize the promise immediately rather than setting it to null. However, if we do this without using a static variable, the API calls will occur twice since both the results and steps datasources extend this base class. Based on our discussion, I have updated the implementation to use a private static promise with public getters and setters to ensure the API call happens only once and is shared across subclasses.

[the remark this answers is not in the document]
ranjgith-ni: @Ahalya-ni - this has not been addressed yet

ranjgith-ni:
Requested changes based on the reply on [the previous comment](https://github.com/ni/systemlink-grafana-plugins/pull/213#discussion_r2111391494)

src/datasources/results/ResultsDataSourceBase.ts
Ahalya-ni: Thank you for clarifying the intended approach and providing the detailed example. I’ve updated the implementation accordingly to follow this pattern for both workspaces and part numbers.

ranjgith-ni (approved):

PR #226, opened 2025-05-30
feat(workspace-util): create workspace reusable to remove duplicate methods of loading workspace in data sources

priya-ni:
# Pull Request

## ✅ Checklist

### DIFF
--- a/src/core/workspace.utils.ts
+++ b/src/core/workspace.utils.ts
@@ -1,0 +1,48 @@
+import { DataSourceInstanceSettings } from "@grafana/data";
+import { BackendSrv } from "@grafana/runtime";
+import { Workspace } from "./types";
+
+export class WorkspaceUtils {
+    readonly workspacesCache = new Map<string, Workspace>([]);
+    workspacesPromise: Promise<Map<string, Workspace>> | null = null;
+
+    queryWorkspacesUrl = `${this.instanceSettings.url}/niauth/v1/auth`;
+
+    constructor(
+        readonly instanceSettings: DataSourceInstanceSettings,
+        readonly backendSrv: BackendSrv
+    ) {
+        this.loadWorkspaces();
+    }
+
+    private static Workspaces: Workspace[] | null = null;
+
+    async loadWorkspaces(): Promise<Map<string, Workspace>> {
+        if (this.workspacesCache.size > 0) {
+            return this.workspacesCache;
+        }
+
+        if (this.workspacesPromise) {
+            return this.workspacesPromise;
+        }
+
+        this.workspacesPromise = this.getWorkspaces().then(workspaces => {
+            if (workspaces) {
+                workspaces.forEach(workspace => this.workspacesCache.set(workspace.id, workspace));
+            }
+            return this.workspacesCache;
+        });
+
+        return this.workspacesPromise;
+    }
+
+    async getWorkspaces(): Promise<Workspace[]> {
+        if (WorkspaceUtils.Workspaces) {
+            return WorkspaceUtils.Workspaces;
+        }
+
+        const response = await this.backendSrv.get<{ workspaces: Workspace[] }>(this.queryWorkspacesUrl);
+
+        return (WorkspaceUtils.Workspaces = response.workspaces);
+    }
+}
--- a/src/datasources/test-plans/TestPlansDataSource.ts
+++ b/src/datasources/test-plans/TestPlansDataSource.ts
@@ -2,10 +2,11 @@ import { DataFrameDTO, DataQueryRequest, DataSourceInstanceSettings, FieldType, 
 import { BackendSrv, TemplateSrv, getBackendSrv, getTemplateSrv } from '@grafana/runtime';
 import { DataSourceBase } from 'core/DataSourceBase';
 import { OrderByOptions, OutputType, Projections, Properties, PropertiesProjectionMap, QueryTestPlansResponse, TestPlanResponseProperties, TestPlansQuery, TestPlansVariableQuery } from './types';
-import { queryInBatches } from 'core/utils';
+import { getWorkspaceName, queryInBatches } from 'core/utils';
 import { QueryResponse } from 'core/types';
 import { isTimeField } from './utils';
 import { QUERY_TEST_PLANS_MAX_TAKE, QUERY_TEST_PLANS_REQUEST_PER_SECOND } from './constants/QueryTestPlans.constants';
+import { WorkspaceUtils } from 'core/workspace.utils';
 
 export class TestPlansDataSource extends DataSourceBase<TestPlansQuery> {
   constructor(
@@ -14,10 +15,12 @@ export class TestPlansDataSource extends DataSourceBase<TestPlansQuery> {
     readonly templateSrv: TemplateSrv = getTemplateSrv()
   ) {
     super(instanceSettings, backendSrv, templateSrv);
+    this.workspaceUtils = new WorkspaceUtils(this.instanceSettings, this.backendSrv);
   }
 
   baseUrl = `${this.instanceSettings.url}/niworkorder/v1`;
   queryTestPlansUrl = `${this.baseUrl}/query-testplans`;
+  workspaceUtils: WorkspaceUtils;
 
   defaultQuery = {
     outputType: OutputType.Properties,
@@ -58,10 +61,19 @@ export class TestPlansDataSource extends DataSourceBase<TestPlansQuery> {
           const fieldType = isTimeField(field)
             ? FieldType.time
             : FieldType.string;
-          const fieldValues = testPlans
+          const values = testPlans
             .map(data => data[field as unknown as keyof TestPlanResponseProperties] as string);
 
           // TODO: AB#3133188 Add support for other field mapping
+          const fieldValues = values.map(value => {
+            switch (field) {
+              case PropertiesProjectionMap.WORKSPACE.field[0]:
+                const workspace = this.workspaceUtils.workspacesCache.get(value);
+                return workspace ? getWorkspaceName([workspace], value) : value;
+              default:
+                return value == null ? '' : value;
+            }
+          });
 
           return {
             name: data.label,
--- a/src/core/workspace.utils.test.ts
+++ b/src/core/workspace.utils.test.ts
@@ -1,0 +1,52 @@
+import { WorkspaceUtils } from './workspace.utils';
+import { DataSourceInstanceSettings } from '@grafana/data';
+import { BackendSrv } from '@grafana/runtime';
+import { Workspace } from './types';
+
+describe('WorkspaceUtils', () => {
+    let instanceSettings: DataSourceInstanceSettings;
+    let backendSrv: BackendSrv;
+    let workspaceUtils: WorkspaceUtils;
+
+    const mockWorkspaces: Workspace[] = [
+        { id: '1', name: 'Workspace 1', default: true, enabled: true },
+        { id: '2', name: 'Workspace 2', default: false, enabled: true }
+    ];
+
+    beforeEach(() => {
+        instanceSettings = { url: 'http://localhost' } as DataSourceInstanceSettings;
+        backendSrv = {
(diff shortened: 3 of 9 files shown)


src/core/workspace.utils.ts
```diff
@@ -0,0 +1,48 @@
...
+
+        if (this.workspacesPromise) {
+            return this.workspacesPromise;
+        }
+
+        this.workspacesPromise = this.getWorkspaces().then(workspaces => {
+            if (workspaces) {
+                workspaces.forEach(workspace => this.workspacesCache.set(workspace.id, workspace));
+            }
+            return this.workspacesCache;
```
atmgrifter00, line 33: I don't understand the purpose of this field. You effectively have two caches now `workspacesCache` (which nothing uses from what I can tell), and `Workspaces`. I suppose you have that cache because `loadWorkspaces` is public, but then the question is why is it? We're already calling it from the .ctor, and subsequent calls should do nothing.

src/core/workspace.utils.ts
```diff
@@ -0,0 +1,48 @@
+import { DataSourceInstanceSettings } from "@grafana/data";
+import { BackendSrv } from "@grafana/runtime";
+import { Workspace } from "./types";
+
+export class WorkspaceUtils {
```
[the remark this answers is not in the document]
priya-ni: I moved that to `shared` folder.

src/core/workspace.utils.ts
priya-ni: I have updated the class to have private methods and removed the static `Workspaces` property, and `workspacesCache ` is needed for the query editor & data source components to get the values.

src/datasources/test-plans/components/TestPlansVariableQueryEditor.test.tsx
```diff
@@ -11,6 +11,13 @@ const mockOnChange = jest.fn();
 const mockOnRunQuery = jest.fn();
 const mockDatasource = {
   prepareQuery: jest.fn((query: TestPlansVariableQuery) => query),
+  workspaceUtils: {
```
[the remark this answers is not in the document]
priya-ni: added tests.

src/shared/Workspaces.ts
```diff
@@ -0,0 +1,41 @@
+import { DataSourceInstanceSettings } from "@grafana/data";
+import { BackendSrv } from "@grafana/runtime";
+import { Workspace } from "core/types";
+
+export class Workspaces {
+    readonly workspacesCache = new Map<string, Workspace>([]);
+    workspacesPromise: Promise<Map<string, Workspace>> | null = null;
+
+    private readonly queryWorkspacesUrl = `${this.instanceSettings.url}/niauth/v1/auth`;
```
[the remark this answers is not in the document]
priya-ni: updated as suggested.

src/shared/workspace.utils.ts
```diff
@@ -0,0 +1,46 @@
...
+    constructor(
+        readonly instanceSettings: DataSourceInstanceSettings,
+        readonly backendSrv: BackendSrv
+    ) {
+        this.loadWorkspaces();
+    }
+
+    get workspacesCache(): Promise<Map<string, Workspace>> {
+        return this._workspacesCache ?? this.loadWorkspaces();
+    }
```
rajsite, line 19: Exposing an async resolved property named "cache" feels like a name that is exposing implementation details. It's an unexpected name.  Could we instead make an async method that resolves to workspaces? Something like `async getWorkspaces(): Promise<Map<string, Workspace>>`

src/shared/workspace.utils.ts
```diff
@@ -0,0 +1,46 @@
+import { DataSourceInstanceSettings } from "@grafana/data";
```
rajsite, line 1: Ideally everywhere but particularly for the `*.utils.ts` files that are being added, lets avoid mixing async/await syntax and manual `.then()` and `.catch()` syntax. We should prefer only using async/await syntax.

In practice this means every function that returns a `Promise` should be async and have an await step statement inside it. We should not see `.then()` or `.catch()` invocations.

In the NI JavaScript Styleguide we have [enabled lint rules](https://github.com/ni/javascript-styleguide/commit/2fbb47b12aa37cb8237828343e2f0c36c02b815a) to help enforce that pattern.

We should avoid introducing mixed usage for the current files in these PRs and consider follow-on tech debt to enable those rules for the repo as a whole.

src/datasources/test-plans/components/TestPlansVariableQueryEditor.test.tsx
```diff
@@ -33,31 +41,35 @@ describe('TestPlansVariableQueryEditor', () => {
   }
 
   it('should render default query', async () => {
-    const container = renderElement();
+    await act(async () => {
+      renderElement();
+    });
```
rajsite, line 46: I wasn't familiar with it but looks like [wrapping component rendering in React with `act` calls](https://react.dev/reference/react/act#rendering-components-in-tests) is an established pattern. And now components have a consistent async `useEffect` to wait for.

Instead of wrapping every call to our custom `renderElement` in an `act` statement can we instead make `renderElement` async and perform the `act` consistently? Seems like it would help with making sure it's used consistently.

src/shared/workspace.utils.ts
priya-ni: I updated the file with await/async and `try catch` for error handling instead of mixing it with `.then() & .catch()`.

For updating the other places, I created a TD [Technical Debt 3152639](https://ni.visualstudio.com/DevCentral/_workitems/edit/3152639): Enable lint rule "require-await": "error" for systemlink-grafana-plugins  repo.

src/shared/workspace.utils.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    constructor(
+        readonly instanceSettings: DataSourceInstanceSettings,
+        readonly backendSrv: BackendSrv
+    ) {
+        this.loadWorkspaces();
+    }
+
+    async getWorkspaces(): Promise<Map<string, Workspace>> {
+        return this._workspacesCache ?? await this.loadWorkspaces();
```
atmgrifter00, line 18: Is there any concern about the cache becoming stale? If a workspace is added while a grafana dashboard is open, currently, this utility will never find it. Is it appropriate to have the cache get refreshed after some elapsed time?

ranjgith-ni: @rajsite - The notion of this change is to enable some reusability between data sources on how each of the data sources cache these values. This change consolidates the caching logic into a shared class so every data source can reuse it.

While the update is mainly a refactor, I understand your concerns and there is a lot of improvement in how we perform http calls and how important it is for us to get the answers to these questions.

Can we continue having this conversation in the tech debt? [which we (@priya-ni and I) are targeting in the next iteration i07]

src/shared/workspace.utils.ts
```diff
@@ -0,0 +1,45 @@
+import { DataSourceInstanceSettings } from "@grafana/data";
+import { BackendSrv } from "@grafana/runtime";
+import { Workspace } from "core/types";
+
+export class WorkspaceUtils {
+    private _workspacesCache: Promise<Map<string, Workspace>> | null = null;
```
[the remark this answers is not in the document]
atmgrifter00: Also, is there a reason for this cache to not be `static`?

src/shared/workspace.utils.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    private readonly queryWorkspacesUrl = `${this.instanceSettings.url}/niauth/v1/auth`;
+
+    constructor(
+        readonly instanceSettings: DataSourceInstanceSettings,
+        readonly backendSrv: BackendSrv
+    ) {
+        this.loadWorkspaces();
```
atmgrifter00, line 14: As is now being done in the ["Users" PR](https://github.com/ni/systemlink-grafana-plugins/pull/227/files#diff-a85aee6a56598cda66773695e7df8c523bdd116ff7715e287fd7c4cfec30b061R31), I don't think there is a strong reason to do this in the .ctor.

src/shared/workspace.utils.ts
```diff
@@ -0,0 +1,45 @@
...
+            }
+            this._workspacesCache = Promise.resolve(workspaceMap);
+            return workspaceMap;
+        } catch (error) {
+            console.error('Error in loading workspaces:', error);
+            const emptyMap = new Map<string, Workspace>();
+            this._workspacesCache = Promise.resolve(emptyMap);
+            return emptyMap;
+        }
+    }
```
rajsite, line 39: We don't need to manually manage the workspacesCache promise, calling loadWorkspaces itself is the promise. You can see that in the function signature that returns `Promise<Map<string, Workspace>>`.

We can just set workspacesCache to the result of invoking loadWorkspaces, we don't need all the manual Promise management. For example (and you can try it on [jsbin](https://jsbin.com/yiquxuv/1/edit?js,console)):

```
class MyCache {
    async getItems() {
        if(!this._itemsPromise) {
            this._itemsPromise = /*note: not using await here*/ this.fetchItems();
        }
        return await this._itemsPromise;
    }
    async fetchItems () {
        await new Promise(resolve => setTimeout(resolve, 200));
        console.log('items fetched!')
        return ['yay!'];
    }
}

const myCache = new MyCache();
console.log('1', myCache._itemsPromise ? 'promise made' : 'promise not made' );
const itemsPromise = myCache.getItems();
console.log('2', myCache._itemsPromise ? 'promise made' : 'promise not made' );
await itemsPromise;
console.log('done');
```

src/datasources/test-plans/TestPlansDataSource.ts
```diff
@@ -58,10 +62,19 @@ export class TestPlansDataSource extends DataSourceBase<TestPlansQuery> {
           const fieldType = isTimeField(field)
             ? FieldType.time
             : FieldType.string;
-          const fieldValues = testPlans
+          const values = testPlans
             .map(data => data[field as unknown as keyof TestPlanResponseProperties] as string);
 
           // TODO: AB#3133188 Add support for other field mapping
+          const fieldValues = values.map(value => {
+            switch (field) {
```
atmgrifter00, line 70: As I mention in [this PR](https://github.com/ni/systemlink-grafana-plugins/pull/234/files#diff-631dbed5ac27a30a485a4539601dd7fde1de40eb47a7309585fadf31ec344203R71), this implementation is at odds with the other and can not be sanely merged.

src/datasources/test-plans/TestPlansDataSource.ts
```diff
@@ -58,10 +62,19 @@ export class TestPlansDataSource extends DataSourceBase<TestPlansQuery> {
...
+              case PropertiesProjectionMap.WORKSPACE.field[0]:
```
atmgrifter00, line 71: Specifying the `case` value in this way is not clear. Is there a more readable option?

src/shared/workspace.utils.ts
priya-ni: Removed loading workspaces in the constructor.

src/shared/workspace.utils.ts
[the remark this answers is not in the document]
priya-ni: Moved the early return statement to `getWorkspaces()` and updated `LoadWorkspaces()` to call and return the response.

src/shared/workspace.utils.ts
priya-ni: Applied the same pattern to `_workspacesCache`.

src/datasources/test-plans/TestPlansDataSource.ts
priya-ni: Updated the case concerning `label` for the property.

src/datasources/test-plans/TestPlansDataSource.ts
priya-ni: Updated the implementation to use `label` instead of field. So that it will return a proper response based for the user-selected `Properties` from the UI.

atmgrifter00 (approved):

rajsite (approved):

PR #227, opened 2025-05-30
feat(shared): Add users class with utility methods to load users

manisha-ni:
# Pull Request

## 🤨 Rationale

To create a `Users` class which can be used by datasources to query users

## 👩‍💻 Implementation

- Added a method in utils to query in batches until continuation token is null
- Added a users class which loads users and stores cache of users and users map
- Added methods to get fullname and fullname with email of users

## 🧪 Testing

- Added unit tests

## ✅ Checklist

### DIFF
--- a/src/core/utils.ts
+++ b/src/core/utils.ts
@@ -176,6 +176,46 @@ export async function queryInBatches<T>(
   };
 }
 
+/**
+ * Executes a query function repeatedly until continuation token is null i.e. all data is retrieved, adhering to the specified
+ * batch query configuration for maximum requests per second and items per request.
+ *
+ * @template T - The type of the data being queried.
+ * @param queryRecord - A function that performs the query. It takes the maximum number of items
+ *                      to retrieve (`take`) and an optional continuation token, and returns a
+ *                      promise that resolves to a `QueryResponse<T>`.
+ * @param config - The batch query configuration, including:
+ *   - `maxTakePerRequest`: The maximum number of items to retrieve per request.
+ *   - `requestsPerSecond`: The maximum number of requests to make per second.
+ * @returns A promise that containing all retrieved data
+ */
+export async function queryUntilComplete<T>(
+  queryRecord: (take: number, continuationToken?: string) => Promise<QueryResponse<T>>,
+  { maxTakePerRequest, requestsPerSecond }: BatchQueryConfig,
+): Promise<QueryResponse<T>> {
+  const data: T[] = [];
+  let continuationToken: string | undefined;
+
+  do {
+    const start = Date.now();
+    for (let i = 0; i < requestsPerSecond; i++) {
+      const response: QueryResponse<T> = await queryRecord(maxTakePerRequest, continuationToken);
+      data.push(...response.data);
+      continuationToken = response.continuationToken;
+
+      if (!continuationToken) {
+        break;
+      }
+    }
+    const elapsed = Date.now() - start;
+    if (continuationToken && elapsed < 1000) {
+      await delay(1000 - elapsed);
+    }
+  } while(continuationToken);
+
+  return { data };
+}
+
 async function delay(timeout: number): Promise<void> {
   return new Promise(resolve => setTimeout(resolve, timeout));
 }
--- a/src/shared/Users.ts
+++ b/src/shared/Users.ts
@@ -1,0 +1,125 @@
+import { DataSourceInstanceSettings } from '@grafana/data';
+import { BackendSrv } from '@grafana/runtime';
+import { queryUntilComplete } from 'core/utils';
+import { QueryResponse } from 'core/types';
+import { User, QueryUsersRequest, QueryUsersResponse } from './types/QueryUsers.types';
+import { QUERY_USERS_MAX_TAKE, QUERY_USERS_REQUEST_PER_SECOND } from './constants/Users.constants';
+
+/**
+ * Represents a utility class for managing and querying user data.
+ */
+export class Users {
+  /**
+   * A cached promise that resolves to an array of users.
+   * This cache is used to avoid redundant user data fetches.
+   */
+  private static _usersCache: Promise<User[]> | null = null;
+
+  /**
+   * A cached promise that resolves to a map of user IDs to their full names.
+   * This cache is used to optimize user name lookups.
+   */
+  private static _usersMapCache: Promise<Map<string, string>> | null = null;
+
+  
+  /**
+   * Retrieves the cached promise for the list of users.
+   * If the cache is not initialized, it triggers the loading of users.
+   */
+  get usersCache(): Promise<User[]> {
+    return this.loadUsers();
+  }
+
+  /**
+   * Retrieves the cached promise for the map of user IDs to full names.
+   * If the cache is not initialized, it generates the map from the user list.
+   */
+  get usersMapCache(): Promise<Map<string, string>> {
+    if (Users._usersMapCache) {
+      return Users._usersMapCache;
+    }
+    return this.usersCache.then(users => {
+      const userMap = new Map<string, string>();
+      users.forEach(user => {
+        const fullName = Users.getUserFullName(user);
+        userMap.set(user.id, fullName);
+      });
+      return userMap;
+    });
+  }
+
+  constructor(readonly instanceSettings: DataSourceInstanceSettings, readonly backendSrv: BackendSrv) {
+    this.loadUsers();
+  }
+
+  /**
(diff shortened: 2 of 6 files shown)


priya-ni (approved):

src/core/utils.ts
```diff
@@ -176,6 +176,46 @@ export async function queryInBatches<T>(
...
+ * @template T - The type of the data being queried.
+ * @param queryRecord - A function that performs the query. It takes the maximum number of items
+ *                      to retrieve (`take`) and an optional continuation token, and returns a
+ *                      promise that resolves to a `QueryResponse<T>`.
+ * @param config - The batch query configuration, including:
+ *   - `maxTakePerRequest`: The maximum number of items to retrieve per request.
+ *   - `requestsPerSecond`: The maximum number of requests to make per second.
+ * @returns A promise that containing all retrieved data
+ */
+export async function queryUntilComplete<T>(
```
atmgrifter00, line 192: The primary concerns I have are a) whether or not there should be explicit error handling in this function, and b) this function allows for an unbounded number of items to be retrieved. While it looks like this utility method was put in place initially to retrieve users, which shouldn't have a huge number, this mechanism is still meant to be general-purpose (rightly so), but as a result can certainly run into issues when used for certain things.

src/shared/Users.ts
```diff
@@ -0,0 +1,125 @@
...
+      users.forEach(user => {
+        const fullName = Users.getUserFullName(user);
+        userMap.set(user.id, fullName);
+      });
+      return userMap;
+    });
+  }
+
+  constructor(readonly instanceSettings: DataSourceInstanceSettings, readonly backendSrv: BackendSrv) {
+    this.loadUsers();
```
atmgrifter00, line 52: Why is this needed with `usersCache()` performing the same operation?

src/shared/Users.ts
```diff
@@ -0,0 +1,125 @@
...
+  public static getUserNameAndEmail(user: User): string {
+    return `${user?.firstName ?? ''} ${user?.lastName ?? ''} (${user?.email ?? ''})`;
+  }
+
+  /**
+   * Loads the list of users from the backend or returns the cached users if available.
+   * This method ensures that user data is fetched only once and reused across the application.
+   * @returns A promise that resolves to an array of users.
+   */
+  private async loadUsers(): Promise<User[]> {
```
atmgrifter00, line 78: I guess I don't understand having both this _and_ `usersCache`. It seems you could just move this implementation into `usersCache`, and delete this method entirely.

src/shared/Users.ts
```diff
@@ -0,0 +1,125 @@
...
+   * This cache is used to optimize user name lookups.
+   */
+  private static _usersMapCache: Promise<Map<string, string>> | null = null;
+
+  
+  /**
+   * Retrieves the cached promise for the list of users.
+   * If the cache is not initialized, it triggers the loading of users.
+   */
+  get usersCache(): Promise<User[]> {
```
atmgrifter00, line 29: I suspect this is meant to be public. If so, please demark it with `public`. Additionally, as a public method, the fact that this uses a cache is an implementation detail, and unnecessary to be part of the name semantically. One alternative would simply be `users`.

src/shared/Users.ts
```diff
@@ -0,0 +1,125 @@
...
+   * A cached promise that resolves to an array of users.
+   * This cache is used to avoid redundant user data fetches.
+   */
+  private static _usersCache: Promise<User[]> | null = null;
+
+  /**
+   * A cached promise that resolves to a map of user IDs to their full names.
+   * This cache is used to optimize user name lookups.
+   */
+  private static _usersMapCache: Promise<Map<string, string>> | null = null;
```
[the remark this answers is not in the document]
manisha-ni: UsersMap is used when [processing workorders data and map user id to their name](https://github.com/ni/systemlink-grafana-plugins/pull/228/files#diff-47ee56625cdb383dfce45154f03b704f5691186bde30d57472b4ed3f9caf6112R149) and [users is to populate the lookup in querybuilder](https://github.com/ni/systemlink-grafana-plugins/pull/228/files#diff-47ee56625cdb383dfce45154f03b704f5691186bde30d57472b4ed3f9caf6112R149)

We don't really need a cache for it but since waiting for users to load and converting to map is common usage, I stored it in class

src/shared/Users.ts
manisha-ni: `WorkspaceUtils` have also been moved to use getters so applied you earlier suggestion to mark it public and remove cache from name

src/core/utils.ts
```diff
@@ -176,6 +176,50 @@ export async function queryInBatches<T>(
...
+  queryRecord: (take: number, continuationToken?: string) => Promise<QueryResponse<T>>,
+  { maxTakePerRequest, requestsPerSecond }: BatchQueryConfig
+): Promise<QueryResponse<T>> {
```
rajsite, line 195: @manisha-ni  @rbell517 I don't see this kind of rate limiting in our Angular services, [for example the sl-users-service](https://dev.azure.com/ni/DevCentral/_git/Skyline?path=%2FWeb%2FWorkspaces%2FSystemLinkShared%2Fprojects%2Fsystemlink-lib-angular%2Fsrc%2Fservices%2Fsl-users-service.ts&_a=contents&version=GBmaster).
- Am I missing it and it's represented elsewhere in angular services?
- Is there a reason specifically this service and no other data services need it?

rbell517: We do have cases in the Angular side where we self-throttle ([steps table](https://dev.azure.com/ni/DevCentral/_git/Skyline?path=/Web/Workspaces/TestInsights/src/app/results/services/steps-grid-core/steps-grid-core.service.ts&version=GBmaster&line=138&lineEnd=148&lineStartColumn=1&lineEndColumn=11&lineStyle=plain&_a=contents) for example) but that is a pattern I have been discouraging in favor of retrying rate limited responses using the common HTTP client interceptor. That is generally because any static throttle we apply is ignorant of the actual rate of requests the backend would allow, as that depends on customer configuration and number of service replicas that have spun up. Going as fast as you can then retrying with a backoff allows a more dynamic automatic limiting.

Manisha's point is a good one though, this retry approach might not be as reasonable in Grafana where you have less control over the number of requests a given page is going to make. In the Angular side we have a good idea of how bounded the number of requests an operation might take (though it is often still variable) and how likely it is to hit the rate limit. Even though retries are generally safe, there is still extra cost to the service when you make requests it turns around and rejects.

I do see value in trying to establish a global throttle across all requests make to our APIs instead of trying to do something in each data plugin. We would need to understand if that is workable for all the different services and their different rate limit defaults. In Angular this could be done as part of a root HTTP interceptor and I imagine we could find a similar approach for this code. Can you capture tech debt to investigate this?

src/shared/users.utils.ts
```diff
@@ -0,0 +1,125 @@
...
+  public static getUserFullName(user: User): string {
+    return `${user.firstName} ${user.lastName}`;
+  }
+
+  /**
+   * Generates a string containing the user's full name and email address.
+   * @param user - The user object.
+   * @returns A string in the format "FirstName LastName (Email)".
+   */
+  public static getUserNameAndEmail(user: User): string {
```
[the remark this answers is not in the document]
manisha-ni: The logic for forming full names and combining usernames with emails will be used by WorkOrders, TestPlans, and potentially other datasources in the future. To avoid duplicating this logic across multiple places, it felt better to centralized here for reuse

src/core/utils.ts
manisha-ni: Captured this in [Technical Debt 3152706](https://dev.azure.com/ni/DevCentral/_workitems/edit/3152706): Systemlink Grafana Plugin | Decide and Implement HTTP Client implementation Strategies

src/shared/users.utils.ts
```diff
@@ -0,0 +1,125 @@
...
+   */
+  public async getusersMap(): Promise<Map<string, string>> {
+    if(!UsersUtils._usersMapCache) {
+      const users = await this.getUsers;
+      const userMap = new Map<string, string>();
+      users.forEach(user => {
+        const fullName = UsersUtils.getUserFullName(user);
+        userMap.set(user.id, fullName);
+      });
+      UsersUtils._usersMapCache = Promise.resolve(userMap);
```
atmgrifter00: Please apply the pattern [@rajsite suggested](https://github.com/ni/systemlink-grafana-plugins/pull/226#discussion_r2130632778) here.

atmgrifter00 (approved):

PR #234, opened 2025-06-02
feat(system-utils): create system reusable to remove duplicate methods of loading system alias in data sources

priya-ni:
# Pull Request

## 🤨 Rationale

We need to query the system alias to map the id to its name in test plans data source.
So created a reusable class SystemUtils to provide property & method needed for data sources

## 🧪 Testing

Added unit tests

## ✅ Checklist

### DIFF
--- a/src/core/utils.ts
+++ b/src/core/utils.ts
@@ -176,6 +176,49 @@ export async function queryInBatches<T>(
   };
 }
 
+export async function queryUsingSkip<T>(
+  queryRecord: (take: number, skip: number) => Promise<QueryResponse<T>>,
+  { maxTakePerRequest, requestsPerSecond }: BatchQueryConfig
+): Promise<QueryResponse<T>> {
+  const delay = (ms: number) => new Promise(resolve => setTimeout(resolve, ms));
+
+  let skip = 0;
+  const data: T[] = [];
+  let hasMore = true;
+
+  do {
+    const start = Date.now();
+
+    for (let i = 0; i < requestsPerSecond && hasMore; i++) {
+      try {
+        const response = await queryRecord(maxTakePerRequest, skip);
+        data.push(...response.data);
+
+        if (response.data.length < maxTakePerRequest) {
+          hasMore = false;
+          break;
+        }
+
+        skip += maxTakePerRequest;
+      } catch (error) {
+        console.error(`Error during batch fetch at skip=${skip}:`, error);
+        hasMore = false;
+        break;
+      }
+    }
+
+    const elapsed = Date.now() - start;
+    if (hasMore && elapsed < 1000) {
+      await delay(1000 - elapsed);
+    }
+  } while (hasMore);
+
+  return {
+    data,
+    totalCount: data.length,
+  };
+}
+
 async function delay(timeout: number): Promise<void> {
   return new Promise(resolve => setTimeout(resolve, timeout));
 }
--- a/src/datasources/test-plans/components/TestPlansQueryEditor.tsx
+++ b/src/datasources/test-plans/components/TestPlansQueryEditor.tsx
@@ -1,15 +1,27 @@
-import React, { useCallback, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { QueryEditorProps, SelectableValue } from '@grafana/data';
 import { TestPlansDataSource } from '../TestPlansDataSource';
 import { OrderBy, OutputType, Properties, PropertiesProjectionMap, TestPlansQuery } from '../types';
 import { AutoSizeInput, HorizontalGroup, InlineField, InlineSwitch, MultiSelect, RadioButtonGroup, Select, VerticalGroup } from '@grafana/ui';
 import { validateNumericInput } from 'core/utils';
 import { TestPlansQueryBuilder } from './query-builder/TestPlansQueryBuilder';
+import { systemAlias } from 'shared/types/QuerySystems.types';
 
 type Props = QueryEditorProps<TestPlansDataSource, TestPlansQuery>;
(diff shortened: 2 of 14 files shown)


src/shared/system.utils.ts
```diff
@@ -0,0 +1,77 @@
+import { DataSourceInstanceSettings } from "@grafana/data";
+import { BackendSrv } from "@grafana/runtime";
+import { QUERY_SYSTEMS_ALIAS_PROJECTION, QUERY_SYSTEMS_MAX_TAKE, QUERY_SYSTEMS_REQUEST_PER_SECOND } from "./constants/QuerySystems.constants";
+import { QueryResponse, QuerySystemsResponse } from "core/types";
+import { queryUsingSkip } from "core/utils";
+import { SystemAlias } from "./types/QuerySystems.types";
+
+export class SystemUtils {
+    private _systemAliasCache: Promise<Map<string, SystemAlias>> | null = null;
```
atmgrifter00, line 9: Can this be `static`? Also, this seems functionally equivalent to:
```ts
private _systemAliasCache?: Promise<Map<string, SystemAlias>>
```

i.e., `null` doesn't provide any value.

src/datasources/test-plans/components/query-builder/TestPlansQueryBuilder.tsx
```diff
@@ -2,25 +2,50 @@ import { SlQueryBuilder } from "core/components/SlQueryBuilder/SlQueryBuilder";
...
+                    ...(systemAliasField.lookup?.dataSource || []),
+                    ...systemAliases.map(({ id, alias }) => ({ label: alias ?? '', value: id })),
+                ],
+            },
+        };
+    }, [systemAliases])
+
     useEffect(() => {
-        const updatedFields = TestPlansQueryBuilderStaticFields
+        if (!systemAliasField) {
```
atmgrifter00, line 44: So both this PR, and [this one](https://github.com/ni/systemlink-grafana-plugins/pull/226/files#) are updating this logic in a similar fashion. I'm curious as to the reason for the early return statements. I suspect that while the prior code was written in a way to suggest that the effect would run when `globalVariableOptions` was updated, it was ultimately unnecessary, as nothing in that state affected the results of the effect implementation. If that is accurate, then I would guess the early returns are just corrections of the prior implementation?

src/shared/system.utils.ts
priya-ni: Applied suggestions.

src/core/utils.ts
```diff
@@ -176,6 +176,49 @@ export async function queryInBatches<T>(
   };
 }
 
+export async function queryUsingSkip<T>(
```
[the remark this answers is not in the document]
priya-ni: Added documentation and notes.

src/datasources/test-plans/components/query-builder/TestPlansQueryBuilder.tsx
priya-ni: Yes, the early return statements are likely a correction to the prior implementation, ensuring that the effect only runs when it is meaningful to do so. Also, I removed `globalVariableOptions `, which is not needed now and will be added once we integrate Grafana variables into the query builder.

src/datasources/test-plans/components/TestPlansVariableQueryEditor.tsx
```diff
@@ -14,6 +15,18 @@ export function TestPlansVariableQueryEditor({ query, onChange, datasource }: Pr
   const [recordCountInvalidMessage, setRecordCountInvalidMessage] = useState<string>('');
 
 
+
```
atmgrifter00, line 18: I would hope the linter complains about the unnecessary newlines here.

src/shared/system.utils.ts
```diff
@@ -0,0 +1,74 @@
...
+export class SystemUtils {
+    private static _systemAliasCache?: Promise<Map<string, SystemAlias>>;
+
+    private readonly querySystemsUrl = `${this.instanceSettings.url}/nisysmgmt/v1/query-systems`;
+
+    constructor(
+        readonly instanceSettings: DataSourceInstanceSettings,
+        readonly backendSrv: BackendSrv
+    ) {
+        this.loadSystems();
```
atmgrifter00, line 17: Think you can remove this from the .ctor.

src/datasources/test-plans/components/TestPlansVariableQueryEditor.tsx
priya-ni: No, linter didn't catch. So, I updated manually now.
Created task to add the lint rule [Task 3154544](https://ni.visualstudio.com/DevCentral/_workitems/edit/3154544): Add lint rule for unnecessary new line..

src/datasources/test-plans/components/TestPlansQueryEditor.tsx
```diff
@@ -41,20 +53,20 @@ export function TestPlansQueryEditor({ query, onChange, onRunQuery, datasource }
...
+        setRecordCountInvalidMessage(recordCountErrorMessages.greaterOrEqualToZero);
+        break;
+      case value > TAKE_LIMIT:
+        setRecordCountInvalidMessage(recordCountErrorMessages.lessOrEqualToTenThousand);
+        break;
+      default:
+        setRecordCountInvalidMessage('');
+        handleQueryChange({ ...query, recordCount: value });
+        break;
+    }
```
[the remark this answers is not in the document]
priya-ni: Applied suggestion.

rajsite (approved):

ushakanagaraj-ni (approved):

PR #242, opened 2025-06-04
feat(test-plan): show fixture and dut name for ids

priya-ni:
# Pull Request

## 🤨 Rationale

In the Test plan data source response, show the fixture names & dut names for their IDs

## 👩‍💻 Implementation
- Created `asset.utils` class to query the assets based on the Ids with batching
- Map the names to fixtureIds & dutIds

## 🧪 Testing

Added unit tests

## ✅ Checklist


src/datasources/test-plans/asset.utils.ts
ranjgith-ni, line 1: Essentially assets http methods should be within the assets data source and be made reusable to be used for other data sources. Since we are rethinking how the http methods are structured in Grafana plugins, I am okay with allowing this for now. But this is something for us to keep in mind during the refactor

ranjgith-ni (approved):

PR #252, opened 2025-06-06
feat(products, results): add validation for property selection

richie-ni:
# Pull Request

## 🤨 Rationale

This pull request introduces validation in the `Properties` element  to ensure that at least one property is selected in the query editors across multiple components. 

### Validation Logic Updates:
* Added a new state variable `isPropertiesValid` in `ProductsQueryEditor`, `QueryResultsEditor`, and `QueryStepsEditor` to track whether at least one property is selected. This state is updated in the `onPropertiesChange` function for each component.

### UI Enhancements:
* Modified the `InlineField` in `ProductsQueryEditor`, `QueryResultsEditor`, and `QueryStepsEditor` to display an error message and mark the field as invalid when no properties are selected. The error message reads: "At least one property must be selected to display data." 

## 🧪 Testing

* Added test cases to `ProductsQueryEditor.test.ts`, `QueryResultsEditor.test.tsx`, and `QueryStepsEditor.test.tsx` to verify that the error message is displayed when all properties are removed. These tests simulate property removal and check for the presence of the error message in the DOM.

## ✅ Checklist


src/datasources/products/components/ProductsQueryEditor.tsx
```diff
@@ -80,7 +82,12 @@ export function ProductsQueryEditor({ query, onChange, onRunQuery, datasource }:
     <>
       <HorizontalGroup align="flex-start">
         <VerticalGroup>
-          <InlineField label="Properties" labelWidth={18} tooltip={tooltips.properties}>
+          <InlineField 
+            label="Properties"
+            labelWidth={18}
+            tooltip={tooltips.properties}
+            invalid={!isPropertiesValid}
+            error='At least one property must be selected to display data.'>
```
richie-ni, line 90: @WillCawley , 
I’ve added an error message to the Properties field to notify the user when it’s left empty, since it’s a required field. Could you please review it : 

Error Message : `At least one property must be selected to display data.`
image

Note: The same error message is used for the other Properties element in Results and steps query editor.

ranjgith-ni (approved):

src/datasources/products/components/ProductsQueryEditor.tsx
WillCawley: Hey @richie-ni. Similar to the other string, my suggestion is to change the sentence from passive to active voice. Here are some ideas to choose from:
-	This field requires that you select at least one property to display your data.
-	You must select at least one property.
-	To display your data, select at least one property.

richie-ni: Updated to `You must select at least one property.`
image

PR #264, opened 2025-06-10
feat(results): rename resultsFilter to resultFilter

ranjgith-ni:
# Pull Request

## 🤨 Rationale

Currently the results filter provided to the [query-steps](https://dev-api.lifecyclesolutions.ni.com/niapis/?urls.primaryName=Test+Monitor#/steps/query-steps-v2) is provided as `resultsFilter` instead of `resutlFilter` due to which the respective filter is not being applied

## 👩‍💻 Implementation

- updated the filter to `resultFilter` so that the filter will be applied as expected

## 🧪 Testing

- updated the existing unit tests

## ✅ Checklist


richie-ni:
@ranjgith-ni ,
Verified the Fix.

PR #265, opened 2025-06-10
feat(results) : Add Result Id lookup into Results Query builder

richie-ni:
# Pull Request

## ✅ Checklist

### DIFF
--- a/src/datasources/results/components/query-builders/query-results/ResultsQueryBuilder.test.tsx
+++ b/src/datasources/results/components/query-builders/query-results/ResultsQueryBuilder.test.tsx
@@ -10,9 +10,10 @@ describe('ResultsQueryBuilder', () => {
     const containerClass = 'smart-filter-group-condition-container';
     const workspace = { id: '1', name: 'Selected workspace' } as Workspace;
     const status = ['PASSED', 'FAILED'];
+    const resultIds = ['result1', 'result2'];
 
     function renderElement(workspaces: Workspace[] | null, status: string[], filter: string, globalVariableOptions: QueryBuilderOption[] = []) {
-      reactNode = React.createElement(ResultsQueryBuilder, { filter, workspaces, status, globalVariableOptions, onChange: jest.fn(), });
+      reactNode = React.createElement(ResultsQueryBuilder, { filter, workspaces, resultIds, status, globalVariableOptions, onChange: jest.fn(), });
       const renderResult = render(reactNode);
       return {
         renderResult,
--- a/src/datasources/results/query-handlers/query-results/QueryResultsDataSource.ts
+++ b/src/datasources/results/query-handlers/query-results/QueryResultsDataSource.ts
@@ -12,6 +12,9 @@ export class QueryResultsDataSource extends ResultsDataSourceBase {
   queryResultsUrl = this.baseUrl + '/v2/query-results';
 
   defaultQuery = defaultResultsQuery;
+  resultId: string[] = [];
+  resultIdChangeCallback?: () => void;
+  previousPartNumberQuery: string[] | undefined = [];
 
   async queryResults(
     filter?: string,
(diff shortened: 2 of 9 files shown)


src/datasources/results/query-handlers/query-results/QueryResultsDataSource.ts
```diff
@@ -99,11 +104,29 @@ export class QueryResultsDataSource extends ResultsDataSourceBase {
...
+  }
+
+  setResultIdChangeCallback(callback: () => void) {
+    this.resultIdChangeCallback = callback;
+  }
+
+  private buildPartNumbersQuery(scopedVars: ScopedVars, partNumberQuery: string[]): string {
+    const partNumberFilter = partNumberQuery.length > 0
+      ? `(${this.buildQueryWithOrOperator(ResultsQueryBuilderFieldNames.PART_NUMBER, partNumberQuery)})`
+      : '';
```
Ahalya-ni, line 121: Can we pass undefined instead of empty string to filter for `query-result-values` API when there is no partnumber provided?

src/datasources/results/query-handlers/query-results/QueryResultsDataSource.ts
```diff
@@ -151,6 +176,20 @@ export class QueryResultsDataSource extends ResultsDataSourceBase {
     return [];
   }
 
+  private async loadResultIdsLookupValues(partNumberQuery: string[] | undefined, scopedVars: ScopedVars): Promise<void> {
+    const partNumbersChanged =
```
Ahalya-ni, line 180: This check is somehow bypassing the changes n take field. Meaning if I change value in take field, `query-result-values` API is called to fetch Ids

src/datasources/results/query-handlers/query-steps/QueryStepsDataSource.ts
```diff
@@ -184,6 +191,8 @@ export class QueryStepsDataSource extends ResultsDataSourceBase {
       };
     }
     query.resultsQuery = this.buildResultsQuery(options.scopedVars, query.partNumberQuery, query.resultsQuery);
+
+    await this.loadResultIdLookupValues(query.partNumberQuery, options.scopedVars);
```
Ahalya-ni, line 195: Should the result IDs lookup be populated only after providing the part number? Or should we populate the lookup with all the IDs as we do in results query type

src/datasources/results/components/query-builders/query-results/ResultsQueryBuilder.test.tsx
```diff
@@ -10,9 +10,10 @@ describe('ResultsQueryBuilder', () => {
     const containerClass = 'smart-filter-group-condition-container';
     const workspace = { id: '1', name: 'Selected workspace' } as Workspace;
     const status = ['PASSED', 'FAILED'];
+    const resultIds = ['result1', 'result2'];
 
     function renderElement(workspaces: Workspace[] | null, status: string[], filter: string, globalVariableOptions: QueryBuilderOption[] = []) {
-      reactNode = React.createElement(ResultsQueryBuilder, { filter, workspaces, status, globalVariableOptions, onChange: jest.fn(), });
+      reactNode = React.createElement(ResultsQueryBuilder, { filter, workspaces, resultIds, status, globalVariableOptions, onChange: jest.fn(), });
```
Ahalya-ni, line 16: Can you add test cases to verify 
1. the result ID selection in query builder as we do for workspace or step path?
2. should not set result ID field when it is in null (initial) state? [Reference to workspace](https://github.com/ni/systemlink-grafana-plugins/blob/6781c833cf3e35fa20a02e910ef7c805898ded7f/src/datasources/results/components/query-builders/query-results/ResultsQueryBuilder.test.tsx#L103)

Ahalya-ni (approved):

src/datasources/results/query-handlers/query-results/QueryResultsDataSource.ts
richie-ni: I couldn't reproduce this scenario, could you share more details on the issue.

PR #270, opened 2025-06-11
fix(results): Set "STARTED_AT" as the default option for the order by field

Ahalya-ni:
# Pull Request

## 🤨 Rationale
As part of this [Task 3141110](https://dev.azure.com/ni/DevCentral/_workitems/edit/3141110): Update orderby and useTimeRange filter with indexed fields
this PR removes the order by option from the UI and sets the default value to `STARTED_AT` (indexed field). For the "results" query type, sorting is descending, and for the "steps" query type, sorting is ascending.

## 👩‍💻 Implementation

- Removed orderby and descending controls from UI in both `QueryResultsEditor` and `QueryStepsEditor`.
- Updated default queries for results and steps.
- Updated `QueryResultsDataSource` and `QueryStepsDataSource` use these default values and call the respective APIs accordingly.

## 🧪 Testing

- Added unit test case to verify the functionality and updated the impact test cases.

## ✅ Checklist

### DIFF
--- a/src/datasources/results/defaultQueries.ts
+++ b/src/datasources/results/defaultQueries.ts
@@ -16,7 +16,7 @@ export const defaultResultsQuery: Omit<QueryResults, 'refId'> = {
     ResultsPropertiesOptions.WORKSPACE
   ] as ResultsProperties[],
   orderBy: "STARTED_AT",
-  descending: false,
+  descending: true,
   recordCount: 1000,
   useTimeRange: false,
   useTimeRangeFor: undefined,
@@ -33,7 +33,7 @@ export const defaultStepsQuery: Omit<QuerySteps, 'refId'> = {
     StepsProperties.totalTimeInSeconds
   ] as StepsProperties[],
   orderBy: "STARTED_AT",
-  descending: true,
+  descending: false,
   recordCount: 10_000,
   useTimeRange: false,
   useTimeRangeFor: undefined,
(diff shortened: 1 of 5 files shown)


src/datasources/results/types/QuerySteps.types.ts
```diff
@@ -14,53 +12,8 @@ export interface QuerySteps extends ResultsQuery {
...
-    description: 'Timestamp when the step was last updated',
-  },
-  {
-    value: 'DATA_MODEL',
-    label: 'Data model',
-    description: 'Data model of the step',
-  },
-];
+export const DefaultOrderBy = 'STARTED_AT';
+export const DefaultDescending = false;
```
ranjgith-ni, line 16: Why are we creating constants in a `*.types.ts` file? and why does a constant start with a capital letter?

ranjgith-ni (changes requested):
I would recommend us to update and use the values from the [`defaultQueries.ts`](https://github.com/ni/systemlink-grafana-plugins/blob/main/src/datasources/results/defaultQueries.ts#L18) file

src/datasources/results/types/QuerySteps.types.ts
Ahalya-ni: As we are now using the values from defaultqueries, these constants are not required.

richie-ni (approved):

PR #273, opened 2025-06-11
fix(results): Set "STARTED_AT" as the default option for the useTimeRange field

Ahalya-ni:
# Pull Request

## 🤨 Rationale

As part of this [Task 3141110](https://dev.azure.com/ni/DevCentral/_workitems/edit/3141110): Update orderby and useTimeRange filter with indexed fields,

this PR removes the "filter by" option from the UI and sets "Started" (an indexed field) as the default for both results and steps query types. With this implementation, enabling the useTimeRange toggle will apply the time range filter using "Started" by default.

## 👩‍💻 Implementation

- Removed `filter by` control from `TimeRangeControls.tsx`.
- Updated the `useTimeRangeFor` property in the default queries for both query types to use "Started".
- Removed `UseTimeRangeFor` enum from `types.ts`.

## 🧪 Testing

Added unit test case to verify the functionality and updated the impacted test cases.

## ✅ Checklist


src/datasources/results/ResultsDataSourceBase.ts
```diff
@@ -16,7 +16,6 @@ export abstract class ResultsDataSourceBase extends DataSourceBase<ResultsQuery>
 
   private timeRange: { [key: string]: string } = {
```
ranjgith-ni, line 17: We no longer need this private variable - you could remove this and update the respective filter in `getTimeRangeFilter` method

ranjgith-ni (changes requested):
Remove thee `useTimeRangeFor` property from the `QueryResults` interface and update in all the places accordingly

src/datasources/results/ResultsDataSourceBase.ts
Ahalya-ni: Removed it.

ranjgith-ni (approved):

PR #285, opened 2025-06-13
fix(results): Enable "return count" by default

Ahalya-ni:
# Pull Request

## 🤨 Rationale

To resolve the issue of indefinite `query-steps` API calls in the results query variable editor, ensure that `returnCount` is always set to true, as batching relies on the `totalCount` value.

## 👩‍💻 Implementation

- Updated `queryStepsInBatches` to always call the `query-steps` API with `returnCount` set to `true`.

## 🧪 Testing

- Updated impacted test cases

## ✅ Checklist

### DIFF
--- a/src/datasources/results/query-handlers/query-steps/QueryStepsDataSource.test.ts
+++ b/src/datasources/results/query-handlers/query-steps/QueryStepsDataSource.test.ts
@@ -483,7 +483,6 @@ describe('QueryStepsDataSource', () => {
           100,
           undefined,
           undefined,
-          true
         );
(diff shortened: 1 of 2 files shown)


ranjgith-ni (changes requested):
Enabling return count as true always will put a significant load on the database - If batching relies on `returnCount`, why would we need a returnCount to be available always?

richie-ni:
An alternative fix has been provided in the following PR,  ** https://github.com/ni/systemlink-grafana-plugins/pull/296.,** so closing this PR

PR #288, opened 2025-06-13
fix(workorders,testplans): Show column names when there is no data and show refID in output's column name

manisha-ni:
# Pull Request

## 🤨 Rationale

To show column names when there is no data and to show outputs from the "Total count" output type as it's refID instead of "Total Count"
image
image
image
image

## 👩‍💻 Implementation

- Remove hardcoded value of "Total count" and passed query name
- Remove the check of length on data when returning fields

## 🧪 Testing

- Updated unit tests

## ✅ Checklist

### DIFF
--- a/src/datasources/test-plans/TestPlansDataSource.test.ts
+++ b/src/datasources/test-plans/TestPlansDataSource.test.ts
@@ -181,7 +181,7 @@ describe('runQuery', () => {
     expect(result.fields[1].values).toEqual(['Active', 'Completed']);
   });
 
-  test('returns empty data frame when no test plans are available', async () => {
+  test('should return field without values when no test plans are available', async () => {
     const query = {
       refId: 'A',
       outputType: OutputType.Properties,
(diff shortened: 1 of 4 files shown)


src/datasources/work-orders/WorkOrdersDataSource.test.ts
```diff
@@ -101,9 +101,10 @@ describe('WorkOrdersDataSource', () => {
   });
 
   describe('runQuery', () => {
-    test('should return empty field when no work orders are found', async () => {
+    test('should return field with no data when no work orders are found', async () => {
```
priya-ni, line 104:
```suggestion
    test('should return field without values when no work orders are found', async () => {
```

src/datasources/test-plans/TestPlansDataSource.test.ts
```diff
@@ -195,7 +195,9 @@ describe('runQuery', () => {
 
     const result = await datastore.runQuery(query, mockOptions);
 
-    expect(result.fields).toHaveLength(0);
+    expect(result.fields).toHaveLength(2);
```
priya-ni, line 198: update name of this test

priya-ni (approved):

src/datasources/work-orders/WorkOrdersDataSource.test.ts
manisha-ni: updated

PR #300, opened 2025-06-16
feat(testplans): Add error handling

manisha-ni:
# Pull Request

## 🤨 Rationale

To show floating errors and warnings when API call fails in testplans datasource

## 👩‍💻 Implementation

- Removed catching of error from utils
- Added floating error in editors
- Added methods to catch and parse errors in datasource

## 🧪 Testing

- Manually verified
- Added unit tests

## ✅ Checklist

### DIFF
--- a/src/datasources/test-plans/TestPlansDataSource.ts
+++ b/src/datasources/test-plans/TestPlansDataSource.ts
@@ -1,9 +1,9 @@
-import { DataFrameDTO, DataQueryRequest, DataSourceInstanceSettings, FieldType, LegacyMetricFindQueryOptions, MetricFindValue, TestDataSourceResponse } from '@grafana/data';
+import { AppEvents, DataFrameDTO, DataQueryRequest, DataSourceInstanceSettings, FieldType, LegacyMetricFindQueryOptions, MetricFindValue, TestDataSourceResponse } from '@grafana/data';
 import { BackendSrv, TemplateSrv, getBackendSrv, getTemplateSrv } from '@grafana/runtime';
 import { DataSourceBase } from 'core/DataSourceBase';
 import { Asset, OrderByOptions, OutputType, Projections, Properties, PropertiesProjectionMap, QueryTemplatesResponse, QueryTestPlansResponse, TemplateResponseProperties, TestPlanResponseProperties, TestPlansQuery, TestPlansVariableQuery } from './types';
 import { getWorkspaceName, getVariableOptions, queryInBatches } from 'core/utils';
-import { QueryBuilderOption, QueryResponse } from 'core/types';
+import { QueryBuilderOption, QueryResponse, Workspace } from 'core/types';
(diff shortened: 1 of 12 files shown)


atmgrifter00 (approved):

src/datasources/test-plans/TestPlansDataSource.ts
```diff
@@ -170,6 +176,49 @@ export class TestPlansDataSource extends DataSourceBase<TestPlansQuery> {
     }
   }
 
+  public async loadWorkspaces(): Promise<Map<string, Workspace>> {
```
priya-ni, line 179:
```suggestion
  private async loadWorkspaces(): Promise<Map<string, Workspace>> {
```
Apply to all.

src/shared/product.utils.test.ts
```diff
@@ -47,22 +47,6 @@ describe('ProductUtils', () => {
         expect(result.get('part-number-2')).toEqual(products[1]);
     });
```
priya-ni, line 49: Instead of removing the test, update it to check throw/propagate the error on API failure.

priya-ni (approved):

src/datasources/test-plans/TestPlansDataSource.ts
```diff
@@ -58,10 +64,10 @@ export class TestPlansDataSource extends DataSourceBase<TestPlansQuery> {
...
 
   async runQuery(query: TestPlansQuery, options: DataQueryRequest): Promise<DataFrameDTO> {
-    const workspaces = await this.workspaceUtils.getWorkspaces();
-    const systemAliases = await this.systemUtils.getSystemAliases();
-    const users = await this.usersUtils.getUsers();
-    const products = await this.productUtils.getProductNamesAndPartNumbers();
+    const workspaces = await this.loadWorkspaces();
+    const systemAliases = await this.loadSystemAliases();
+    const users = await this.loadUsers();
+    const products = await this.loadProductNamesAndPartNumbers();
```
ranjgith-ni, line 70: If `runQuery` is the only place where you would do API calls, you could do a try catch finally block and handled the generic error handling instead of writing a bunch of public wrappers.

There is another way to share this implementation - where you could write a [custom hook[(https://react.dev/learn/reusing-logic-with-custom-hooks) called `handleAPIErrors()` (can be named better) and use the hook to either update the error state or to return the error string.

manisha-ni: These methods are also used by editors, that's why they are in public wrappers.

Custom hook can only be used in react functions and hooks and cannot be used inside the class components

manisha-ni: We can think through better handling of API errors in [Technical Debt 3152706](https://dev.azure.com/ni/DevCentral/_workitems/edit/3152706): Systemlink Grafana Plugin | Decide and Implement HTTP Client implementation Strategies

PR #310, opened 2025-06-18
fix(workorders): Handle invalid queries to return empty data

manisha-ni:
# Pull Request

## 🤨 Rationale

To return empty data when user enters invalid queries and to show error message when all properties are removed

## 👩‍💻 Implementation

- Handled invalidity when all properties are removed
- Check validity of take and properties and return empty data if it's invalid
- Use `prepareQuery` in `metricFindQuery` so that initial run query gets populated with default query

## 🧪 Testing

- Manually verified
- Added unit tests

## ✅ Checklist

### DIFF
--- a/src/datasources/work-orders/WorkOrdersDataSource.test.ts
+++ b/src/datasources/work-orders/WorkOrdersDataSource.test.ts
@@ -107,6 +107,7 @@ describe('WorkOrdersDataSource', () => {
         properties: [WorkOrderPropertiesOptions.NAME],
         outputType: OutputType.Properties,
         queryBy: 'filter',
+        take: 1000
       };
 
       jest.spyOn(datastore, 'queryWorkordersData').mockResolvedValue([]);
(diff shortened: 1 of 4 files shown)


src/datasources/work-orders/WorkOrdersDataSource.ts
```diff
@@ -81,20 +87,24 @@ export class WorkOrdersDataSource extends DataSourceBase<WorkOrdersQuery> {
   async metricFindQuery(
     query: WorkOrdersVariableQuery,
     options: LegacyMetricFindQueryOptions
-  ): Promise<MetricFindValue[]> {    
-    const filter = query.queryBy? 
+  ): Promise<MetricFindValue[]> {
+    const variableQuery = this.prepareQuery(query);
+    if (!this.isTakeValid(variableQuery)) {
+      return [];
+    }
```
priya-ni, line 94: this should be handled in the query editor, since take is optional, it can be undefined. Trigger runQuery only when take is valid from component.

priya-ni (changes requested):

priya-ni (approved):

PR #319, opened 2025-06-19
fix(core): remove error message on 504 status code

stefanc18:
# Pull Request

## 🤨 Rationale
https://dev.azure.com/ni/DevCentral/_workitems/edit/3162580

## 👩‍💻 Implementation

Remove the error message from the error thrown when status code is 504. This way we avoid having the entire HTML shown there that is returned by Cloudflare.

## 🧪 Testing

Tested locally with the Systems datasource and used a fake backend. When I returned 504 from the systems request, the error message wasn't shown, when I returned any other status code, it was shown.

## ✅ Checklist

### DIFF
--- a/src/core/DataSourceBase.ts
+++ b/src/core/DataSourceBase.ts
@@ -55,9 +55,14 @@ export abstract class DataSourceBase<TQuery extends DataQuery, TOptions extends 
       if (isFetchError(error)) {
         const fetchError = error as FetchError;
         const statusCode = fetchError.status;
-        const data = fetchError.data;
-        const errorMessage = data.error?.message || JSON.stringify(data);
-        throw new Error(`Request to url "${options.url}" failed with status code: ${statusCode}. Error message: ${errorMessage}`);
+        const genericErrorMessage = `Request to url "${options.url}" failed with status code: ${statusCode}.`;
+        if (statusCode === 504) {
+          throw new Error(genericErrorMessage);
+        } else {
+          const data = fetchError.data;
+          const errorMessage = data.error?.message || JSON.stringify(data);
+          throw new Error(`${genericErrorMessage} Error message: ${errorMessage}`);
+        }
       }
       throw error;
     }


atmgrifter00 (approved):

src/core/DataSourceBase.ts
```diff
@@ -55,9 +55,14 @@ export abstract class DataSourceBase<TQuery extends DataQuery, TOptions extends
       if (isFetchError(error)) {
         const fetchError = error as FetchError;
         const statusCode = fetchError.status;
-        const data = fetchError.data;
-        const errorMessage = data.error?.message || JSON.stringify(data);
-        throw new Error(`Request to url "${options.url}" failed with status code: ${statusCode}. Error message: ${errorMessage}`);
+        const genericErrorMessage = `Request to url "${options.url}" failed with status code: ${statusCode}.`;
+        if (statusCode === 504) {
```
rajsite, line 59: Sounds like the issue is when the response is HTML instead of JSON. Should we focus on the specific status code or do something general like try and parse thee response as JSON and if that fails return a genericErrorMessage?

rajsite (approved):

src/core/DataSourceBase.ts
stefanc18: I tried the more generic approach that you mentioned and found out that `fetchError.data` gives a rather unpredictable output.

For example, if the response can is a JSON, `fetchError.data` is that exact JSON:
image

On the other hand, if the response is a HTML, it doesn't return it anymore, but instead it's still a JSON with some internal format:
image

So it seems some extensive parsing would have to be done in order to achieve this.

Given that a HTML response is an isolated case for the 504 error returned by Cloudflare, as none of the SystemLink services are actually returning HTML on error, I'm leaning towards keeping the implementation as is, based on the status code. Citing also JD's comment from [this work item](https://dev.azure.com/ni/DevCentral/_workitems/edit/3136647): "Cloudflare firewall that sits in front of our internal clusters is going to trap any 504 response from the webserver and render it's own 504 response, which means the browser will get back a 504 with HTML from Cloudflare in the body"

rajsite: Oh wow, that's really weird, I don't think the browser fetch api would behave this way so something strange happening somewhere. Thanks for investigating and the approach seems reasonable 👍

kkerezsi (approved):

PR #334, opened 2025-06-25
fix: add showErrorAlert option to POST requests across data sources

richie-ni:
# Pull Request

## 🤨 Rationale

When an error occurs during a request made using backendSrv.fetch, we see two error messages appear. This happens because the error is being handled both by backendSrv and the internal fetch method. To fix this, we should set showErrorAlert to false in the options passed to backendSrv.

image

## 👩‍💻 Implementation

* Updated the `post` method to include a `showErrorAlert` parameter, defaulting to `true`. This enables callers to suppress error alerts when necessary.

* Modified calls to the `post` method in the products, results, testplans and workorder datasources instances

## 🧪 Testing

N/A

## ✅ Checklist

### DIFF
--- a/src/core/DataSourceBase.ts
+++ b/src/core/DataSourceBase.ts
@@ -72,8 +72,8 @@ export abstract class DataSourceBase<TQuery extends DataQuery, TOptions extends 
     return this.fetch<T>({ method: 'GET', url, params });
   }
 
-  post<T>(url: string, body: Record<string, any>) {
-    return this.fetch<T>({ method: 'POST', url, data: body });
+  post<T>(url: string, body: Record<string, any>, showErrorAlert = true) {
+    return this.fetch<T>({ method: 'POST', url, data: body, showErrorAlert });
   }
 
   static Workspaces: Workspace[];
--- a/src/datasources/products/ProductsDataSource.ts
+++ b/src/datasources/products/ProductsDataSource.ts
@@ -65,14 +65,18 @@ export class ProductsDataSource extends DataSourceBase<ProductQuery> {
     returnCount = false
   ): Promise<QueryProductResponse> {
     try {
-      const response = await this.post<QueryProductResponse>(this.queryProductsUrl, {
-        filter,
-        orderBy,
-        descending,
-        projection,
-        take,
-        returnCount
-      });
+      const response = await this.post<QueryProductResponse>(
+        this.queryProductsUrl,
+        {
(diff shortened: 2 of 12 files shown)


src/datasources/results/query-handlers/query-steps/QueryStepsDataSource.ts
```diff
@@ -83,16 +83,20 @@ export class QueryStepsDataSource extends ResultsDataSourceBase {
...
-        orderBy,
-        descending,
-        projection,
-        take,
-        resultFilter,
-        continuationToken,
-        returnCount,
-      });
+      const response = await this.post<QueryStepsResponse>(
+        `${this.queryStepsUrl}`,
```
ranjgith-ni, line 87:
```suggestion
        this.queryStepsUrl,
```
nit: Not related the current change, but the templating is not needed

ranjgith-ni (approved):

src/core/DataSourceBase.ts
```diff
@@ -72,8 +72,8 @@ export abstract class DataSourceBase<TQuery extends DataQuery, TOptions extends
     return this.fetch<T>({ method: 'GET', url, params });
```
rajsite, line 72: I don't see tests added to the data sources that cover the desired behavior or documentation describing the pattern.

My concern is if additional data sources are added:

- how are authors expected to know that some calls should hide show error and some should not?
- is there documentation describing how it should be validated? Either with tests or manual validation?
- how do I make sure I don't suppress errors unexpectedly?

src/datasources/products/ProductsDataSource.ts
```diff
@@ -65,14 +65,18 @@ export class ProductsDataSource extends DataSourceBase<ProductQuery> {
...
+        this.queryProductsUrl,
+        {
+          filter,
+          orderBy,
+          descending,
+          projection,
+          take,
+          returnCount
+        },
+        false
```
rajsite, line 78: Consider using an options object instead of adding a flag. Someone reviewing this code in the future will not have a clear understanding of what this random `false` flag is doing at first glance. Can express this pretty well with parameter defaults and destructuring assignment

```
post<T>(url: string, body: Record<string, any>, {
    showErrorAlert = true,
} = { 
    showErrorAlert: true
})
```

```
this.post(..., {showErrorAlert: false})
```

rajsite (approved):

PR #336, opened 2025-06-25
fix(DataSourceBase): ensure URL is preserved during fetch retries

richie-ni:
# Pull Request

## 🤨 Rationale

Currently, when a `429 `- Too Many Requests error occurs during a fetch, the subsequent retry uses a modified URL with `dashboardhost` added to it, which results in a `404` - Page Not Found error.

image

Discussion : https://teams.microsoft.com/l/message/19:b0c2c359cadf418ca24bb9a51a67dcea@thread.tacv2/1750699015524?tenantId=eb06985d-06ca-4a17-81da-629ab99f6505&groupId=1e74afef-f47b-4137-ad1b-e550ec9c6bef&parentMessageId=1750699015524&teamName=TM-SystemLink&channelName=Grafana%20Discussions&createdTime=1750699015524

## 👩‍💻 Implementation

- Preserve the  request url to be used in the retry logic

## 🧪 Testing
 Verified the changes by manually deploy it in main dev ( https://github.com/ni/systemlink-grafana-plugins/pull/333)

## ✅ Checklist

### DIFF
--- a/src/datasources/results/query-handlers/query-results/QueryResultsDataSource.test.ts
+++ b/src/datasources/results/query-handlers/query-results/QueryResultsDataSource.test.ts
@@ -105,7 +105,7 @@ describe('QueryResultsDataSource', () => {
           .toThrow('The query to fetch results failed because the requested resource was not found. Please check the query parameters and try again.');
       });
 
-    test('should throw error when API returns 429 status', async () => {        
+    test('should throw error when API returns 429 status', async () => {
         backendServer.fetch
           .calledWith(requestMatching({ url: '/nitestmonitor/v2/query-results' }))
           .mockReturnValue(createFetchError(429));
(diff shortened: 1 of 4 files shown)


src/datasources/results/query-handlers/query-results/QueryResultsDataSource.test.ts
```diff
@@ -106,6 +107,8 @@ describe('QueryResultsDataSource', () => {
       });
 
     test('should throw error when API returns 429 status', async () => {
+        cleanup();
```
rajsite, line 110: Looking at the docs for this function: https://testing-library.com/docs/react-testing-library/api/#cleanup

It seems to indicate that `cleanup` is used for removing state of rendered React trees. I don't see any React trees rendered here. Is it clear why we need to clean React tree rendered in these tests?

If it is needed can we move it to the `afterEach` so that we can clean-up consistently?
Should we make sure to `cleanup` in afterEach for other tests consistently?

ranjgith-s (approved):

ranjgith-ni (approved):

PR #349, opened 2025-07-02
feat(products): update field labels and refactor projection mapping 

richie-ni:
# Pull Request

## 🤨 Rationale

This pull request updates the column headers in the panel to Sentence case and to improve consistency and readability by replacing hardcoded property names with a centralized lookups (`productsProjectionLabelLookup`). 

## 👩‍💻 Implementation

- Replaced `ProductPropertiesProjectionMap` with `productsProjectionLabelLookup`, which maps `Properties` enum values to their corresponding labels and projections. Updated the `Properties` enum to use lowercase keys for consistency. 
-  Modified the `MultiSelect` component to use `productsProjectionLabelLookup` for generating property options dynamically.

## 🧪 Testing

- Updated the tests and snapshots

## ✅ Checklist

### DIFF
--- a/src/datasources/products/types.ts
+++ b/src/datasources/products/types.ts
@@ -14,15 +14,15 @@ export interface ProductVariableQuery extends DataQuery {
 }
 
 export enum Properties {
-  id = 'ID',
-  partNumber = 'PART_NUMBER',
-  name = 'NAME',
-  family = 'FAMILY',
-  updatedAt = 'UPDATED_AT',
-  workspace = 'WORKSPACE',
-  keywords = 'KEYWORDS',
-  properties = 'PROPERTIES',
-  fileIds = 'FILE_IDS',
+  id = 'id',
+  partNumber = 'partNumber',
+  name = 'name',
+  family = 'family',
+  updatedAt = 'updatedAt',
+  workspace = 'workspace',
+  keywords = 'keywords',
+  properties = 'properties',
+  fileIds = 'fileIds',
 }
 
 export const PropertiesOptions = {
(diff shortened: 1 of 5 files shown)


src/datasources/products/types.ts
```diff
@@ -37,16 +37,18 @@ export const PropertiesOptions = {
...
+  [Properties.id]: { label: 'Product ID', projection: PropertiesOptions.ID},
+  [Properties.partNumber]: { label: 'Part number', projection: PropertiesOptions.PART_NUMBER },
+  [Properties.name]: { label: 'Product name', projection:  PropertiesOptions.NAME },
+  [Properties.family]: { label: 'Family', projection:  PropertiesOptions.FAMILY },
+  [Properties.updatedAt]: { label: 'Updated at', projection: PropertiesOptions.UPDATEDAT },
+  [Properties.workspace]: { label: 'Workspace', projection: PropertiesOptions.WORKSPACE },
+  [Properties.keywords]: { label: 'Keywords', projection: PropertiesOptions.KEYWORDS },
+  [Properties.properties]: { label: 'Properties', projection: PropertiesOptions.PROPERTIES },
+  [Properties.fileIds]: { label: 'File IDs', projection: PropertiesOptions.FILE_IDS },
 }
```
ranjgith-s, line 52: Since we have the same values between the enum and the constant, we could do something like


```suggestion
export const productsProjectionLabelLookup: Record<Properties, {
  label: string;
  projection: Properties;
}> = {
  [Properties.id]: { label: 'Product ID', projection: Properties.id },
  [Properties.partNumber]: { label: 'Part number', projection: Properties.partNumber },
  [Properties.name]: { label: 'Product name', projection: Properties.name },
  [Properties.family]: { label: 'Family', projection: Properties.family },
  [Properties.updatedAt]: { label: 'Updated at', projection: Properties.updatedAt },
  [Properties.workspace]: { label: 'Workspace', projection: Properties.workspace },
  [Properties.keywords]: { label: 'Keywords', projection: Properties.keywords },
  [Properties.properties]: { label: 'Properties', projection: Properties.properties },
  [Properties.fileIds]: { label: 'File IDs', projection: Properties.fileIds },
};
```

This would improve the typing

src/datasources/products/types.ts
```diff
@@ -14,15 +14,15 @@ export interface ProductVariableQuery extends DataQuery {
...
+  name = 'name',
+  family = 'family',
+  updatedAt = 'updatedAt',
+  workspace = 'workspace',
+  keywords = 'keywords',
+  properties = 'properties',
+  fileIds = 'fileIds',
 }
 
 export const PropertiesOptions = {
```
ranjgith-s, line 28: Eventually we should move away from using this constant and replace it with the enum `Properties` This should improve the typing in all the places it is referenced - For instance, [here](https://github.com/ni/systemlink-grafana-plugins/blob/b9eb809bf12f08839c92b430161abf35e009f64b/src/datasources/products/ProductsDataSource.ts#L40) we are doing an implicit type conversion - this could be avoided if we operate only with the enum as the source of truth and use the map for display labels -- I am okay with taking this change as a followup

ranjgith-s (approved):

PR #351, opened 2025-07-07
fix(asset): hide query options from table combo box when query is hidden

OliverCosma:
# Pull Request

## 🤨 Rationale

Bug: disabled queries still appeared in table combo box and were run, even when they were disabled 

e.g.: 
image

## 👩‍💻 Implementation

Don't run the query when it is set on hidden

## 🧪 Testing

image

+ some unit tests
## ✅ Checklist


kkerezsi (approved):

kkerezsi:
@OliverCosma I think Tags as well as Notebook plugins needs to be checked

src/datasources/asset/data-sources/asset-summary/AssetSummaryDataSource.ts
```diff
@@ -24,8 +24,8 @@ export class AssetSummaryDataSource extends AssetDataSourceBase {
         return this.processSummaryQuery(query as AssetQuery);
     }
 
-    shouldRunQuery(_: AssetQuery): boolean {
-        return true;
+    shouldRunQuery(query: AssetQuery): boolean {
+        return !query?.hide;
```
rajsite: @OliverCosma do you remember why optional chaining was used here? Are there scenarios where `query` was `undefined` that you encountered? I'm trying to make the pattern consistent in #456 and opted to use the pattern shown here consistently but was not sure if that was necessary.

PR #358, opened 2025-07-16
fix(results): Fix step path lookup label in dropdown

Ahalya-ni:
# Pull Request

## 🤨 Rationale

This PR provides the fix for [Bug 3192635](https://dev.azure.com/ni/DevCentral/_workitems/edit/3192635): Grafana Result data source | Inconsistency in displaying the step paths when compared to data spaces

## 👩‍💻 Implementation

- Updated the step path type from `string[]` to `Array<{ label: string; value: string }>`  so that the label displays the hierarchy using `\` as a separator, while the value contains the path with `\n` to be sent to the API.
- Modified `getStepPaths` to return the appropriate label by formatting the path with `\` as the separator.

step path

## 🧪 Testing

- Added new and updated existing unit test cases to verify the functionality.

## ✅ Checklist


src/datasources/results/components/query-builders/query-steps/StepsQueryBuilder.tsx
```diff
@@ -20,7 +20,7 @@ type StepsQueryBuilderProps = QueryBuilderProps &
     filter?: string;
     workspaces: Workspace[] | null;
     stepStatus: string[];
-    stepsPath: string[];
+    stepsPath: Array<{ label: string; value: string; }>;
```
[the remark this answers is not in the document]
Ahalya-ni: Created new type

src/datasources/results/query-handlers/query-steps/QueryStepsDataSource.ts
```diff
@@ -301,11 +301,14 @@ export class QueryStepsDataSource extends ResultsDataSourceBase {
     }
   }
 
-  async getStepPaths(resultsQuery: string): Promise<string[]> {
+  async getStepPaths(resultsQuery: string): Promise<Array<{ label: string; value: string; }>> {
```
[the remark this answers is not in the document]
Ahalya-ni: Substituted

gokuldharan-ni (dismissed):
As we are allowing the user to type in path manually to the field, what do we expect here whether the user should type the path separated by `\` or `\n`? In the current approach when the user manually provide a path separated by `\` it won't work. I would like some clarity here. We can have this discussion in the bug thread. 

I am marking request changes untill we gain clarity on this.

Ahalya-ni:
Since the create step API allows creating a step with a null stepname, which was not handled previously in this PR, I have updated the code to address this case. I am resetting your votes to get a review.

@ranjgith-ni @richie-ni

src/datasources/results/components/query-builders/query-steps/StepsQueryBuilder.tsx
```diff
@@ -101,7 +102,7 @@ export const StepsQueryBuilder: React.FC<StepsQueryBuilderProps> = ({
         ...stepsPathField.lookup,
         dataSource: [
           ...(stepsPathField.lookup?.dataSource || []),
-          ...stepsPath.map(path => ({ label: path, value: path })),
+          ...stepsPath.map(path => ({ label: path.label, value: path.value })),
```
ranjgith-ni, line 105:
```suggestion
          ...stepsPath
```

Since the types are same, can't we just do this instead of `Array.map()`? This would mean there is one less copy of the `stepPath`

src/datasources/results/types/QuerySteps.types.ts
```diff
@@ -178,6 +178,11 @@ export enum StepsPathProperties {
   path = 'path'
 };
 
+export type StepPaths = {
```
ranjgith-ni, line 181:
```suggestion
export type StepPath = {
```
nit:

src/datasources/results/components/query-builders/query-steps/StepsQueryBuilder.test.tsx
```diff
@@ -9,14 +9,17 @@ describe('StepsQueryBuilder', () => {
...
+      {label: 'Another Path', value: 'Another Path'},
+    ]
     const status = ['PASSED', 'FAILED'];
 
     function renderElement(
       filter: string,
       workspaces: Workspace[] | null,
       stepStatus: string[],
-      stepsPath: string[],
+      stepsPath: Array<{ label: string; value: string }>,
```
ranjgith-ni, line 22:
```suggestion
      stepsPath: StepPaths[],
```

Use the defined type instead

src/datasources/results/components/query-builders/query-steps/StepsQueryBuilder.test.tsx
```diff
@@ -9,14 +9,17 @@ describe('StepsQueryBuilder', () => {
 
     const containerClass = 'smart-element smart-query-builder';
     const workspace = { id: '1', name: 'Selected workspace' } as Workspace;
-    const stepsPath = ['path1', 'path2'];
+    const stepsPath = [
```
ranjgith-ni, line 12:
```suggestion
    const stepPaths = [
```

src/datasources/results/query-handlers/query-steps/QueryStepsDataSource.ts
```diff
@@ -301,11 +302,14 @@ export class QueryStepsDataSource extends ResultsDataSourceBase {
...
 
-  async getStepPaths(resultsQuery: string): Promise<string[]> {
+  async getStepPaths(resultsQuery: string): Promise<StepPaths[]> {
     if (resultsQuery) {
       const query = this.transformQuery(resultsQuery, this.resultsComputedDataFields, this.scopedVars);
       const stepPaths = await this.getStepPathsLookupValues(query!);
-      return this.flattenAndDeduplicate(stepPaths);
+      return this.flattenAndDeduplicate(stepPaths).map(path => ({
+        label: path?.replace(/\n/g, '\\') || '',
+        value: path || ''
```
ranjgith-ni, line 311: Shouldn't we filter out the null and empty values of paths? There is no value in showing empty values in the dropdown. 

Also Test Monitor assigns the default step path as `root` and will be auto assigned based on the parent step names

ranjgith-ni (approved):

src/datasources/results/query-handlers/query-steps/QueryStepsDataSource.ts
Ahalya-ni: I have filtered out null and empty strings from the step paths.
However, I noticed that in the create dataspace dropdown, null and empty paths are allowed. I’m not sure if this is intentional or a bug. Attaching a screenshot for reference.
image

In Test Monitor, the root is assigned by default to the property called `parentId`, not the path. The path itself is constructed internally using the step names. If the user provides a step name as null, the path is set to null. If the step name is an empty string, the path will also be an empty string.

richie-ni (approved):

src/datasources/results/query-handlers/query-steps/QueryStepsDataSource.ts
Ahalya-ni: 1. I agree that displaying empty paths is not meaningful.
2. The step names are generated from TestStand, so the likelihood of encountering null or empty strings is very low. Please refer this [discussion](https://ni.visualstudio.com/DevCentral/_workitems/edit/3192635#10130242) for more details.

The previous attached screenshot of the dataspace displaying empty dropdowns in those cases and tagged you to confirm whether this behavior in the dataspace is intentional or a bug. I came across this scenario while referring to dataspace during development.

karthikb-ni: There seems to be no value in showing the empty paths that i can think of, regarding whether the dataspaces is a bug i do think it's a bug yes.

Ahalya-ni: @karthikb-ni can you share the TD work item here for reference?

karthikb-ni: https://dev.azure.com/ni/DevCentral/_workitems/edit/3223630

PR #379, opened 2025-08-04
fix(results): Update tooltip description for use time range control

Ahalya-ni:
# Pull Request

## 🤨 Rationale

This PR contains the fix for the _[Bug 3207133](https://dev.azure.com/ni/DevCentral/_workitems/edit/3207133): Test results data source - Help text for 'use time range' doesn't help understand which field the time range is applied to_

## 👩‍💻 Implementation

- Updated tooltip for use time range control in result and step editor to have contain the startedAt field which is the field the time range is applied.

## 🧪 Testing

- Manually tested

## ✅ Checklist

### DIFF
--- a/src/datasources/results/components/editors/time-range/TimeRangeControls.tsx
+++ b/src/datasources/results/components/editors/time-range/TimeRangeControls.tsx
@@ -24,7 +24,7 @@ export function TimeRangeControls({query, handleQueryChange}: Props) {
 }
 
 const tooltips = {
-  useTimeRange: 'This toggle enables querying within the dashboard time range.',
+  useTimeRange: 'This toggle enables querying by started at within the dashboard time range.',
 };
 
 const labels = {


src/datasources/results/components/editors/time-range/TimeRangeControls.tsx
```diff
@@ -24,7 +24,7 @@ export function TimeRangeControls({query, handleQueryChange}: Props) {
 }
 
 const tooltips = {
-  useTimeRange: 'This toggle enables querying within the dashboard time range.',
+  useTimeRange: 'This toggle enables querying by "started at" field within the dashboard time range.',
```
Ahalya-ni, line 27: If we want the tooltip to mention the specific resource, for example,
`This toggle enables querying results by the ‘startedAt’ field within the dashboard time range` or
`This toggle enables querying steps by the ‘startedAt’ field within the dashboard time range`
we would need to pass the resource name into this component.

Currently, this component only contains a single switch. It was originally extracted as a separate component because we had an additional combobox to select which field the time range would apply to, and it was shared between results and steps. 

Now that it’s just a switch, we could consider removing the `TimeRangeControls` component and placing the switch directly inside the relevant components, and update tooltips separately. This would require updates to a couple of components and their corresponding test files.

richie-ni: i think let's keep it generic

richie-ni (approved):

src/datasources/results/components/editors/time-range/TimeRangeControls.tsx
Ahalya-ni: @WillCawley , Can you review this update on this user facing string?
image

For context: This field and its tooltip are shared between both results and steps query types. If we want to mention the query type (results/steps) in the tooltip as mentioned in the previous comment, some implementation changes will be needed, which @hemkumarrao-ni  will review in detail later.

WillCawley: I don't think we need to specify the query type unless we plan on adding more types in the future.

What does the "started at" field do for the query? Its not clear why the user needs to know this based on the proposed tooltip

Here are some suggestions:

- This toggle enables querying by using the value in the "started at" field within the dashboard time range.
- This toggle enables querying within the dashboard time range. The query uses the value in the “started at" field as a beginning point for the time range.

By the way, why did we change "startedAt" to "started at"? Is 'started at' the actual name of the field? If so, shouldn't it be upper case ('Started At') instead?

WillCawley: @Ahalya-ni What I mean is, will the customer interact with a UI element called "started at"? Or is that purely a filter for the query?

Ahalya-ni: There is no interactive element in the UI labeled "started at". It is the underlying field used for filtering behind the scenes and is included in the query passed to the API.

WillCawley: Discussed with @Ahalya-ni in Teams. Based on that, here is a new suggestion:
- This toggle enables querying a dashboard within a time range. The query uses an internal value ('StartedAt') as the beginning point.

karthikb-ni: I think there is some misunderstanding/miscommunication based on what i see in this comment thread @Ahalya-ni @WillCawley , i would like to clarify a few things to ensure everyone understands and then offer my suggestion.

There is a "Started at" property that the user can select which is then shown in the grafana table.
image, this is internally in the code known as "startedAt" and shown to the user as "Started At".

The user cannot change what time field is used for filtering, so for example the user cannot choose filtering to be done by "updatedAt". So the time field used for filtering is fixed but it does show up in the UI.

So keeping this in mind i think the tooltip should be "This toggle enables filtering the response data to only return those responses where "Started at" falls within the specified dashboard time range"

WillCawley: @karthikb-ni Thank you for the clarification. Here is a slight edit to what you proposed:

- This toggle enables the querying of response data with a filter. This filter only returns responses with a "Started at" property value that falls within a specified dashboard time range.

WillCawley (approved):

hemkumarrao-ni (approved):

PR #387, opened 2025-08-12
feat(asset):  add take support for query variables for ListAssets data source

BlankaOnodi:
# Pull Request

## 🤨 Rationale

This PR adds support for the take parameter in ListAssets data source variable queries, allowing users to limit the number of returned results when defining dashboard variables.
The take value can be configured from the UI, with proper validation to avoid invalid or excessive limits.
A default value is provided via defaultListAssetsVariable.

## 👩‍💻 Implementation

Query handling:
  Added patchListAssetQueryVariable(query) function in AssetDataSource.ts:
    1. Merges the provided query with defaultListAssetsVariable to ensure backward compatibility with existing dashboards.
    2. Default values: filter: "", take: QUERY_LIMIT (1000)
  Integrated patchListAssetsVariableQuery into metricFindQuery for ListAssets variable queries.
  
  Added validation function isTakeValid(take: number | undefined), which is valid if:
    1. take is defined
    2. take >= 0
    3. take <= TAKE_LIMIT. If invalid, returns empty array instead of query results.
  
UI:
  Updated variable editor UI to include an AutoSizeInput for take.
  On change:
   1. Validate input (NaN, < 0, > TAKE_LIMIT are invalid)
   2. Show error message next to field if invalid
   3. If valid, update query .

Backward compatibility:
  Old dashboard variables will work without modification due to the patch function as missing take will be replaced with QUERY_LIMIT 

## 🧪 Testing

Unit testing:
 1. Showing correct error message when take is invalid.
 2. Returning an empty array when take is invalid.
 3. Returning correct results when: take = TAKE_LIMIT
 
Manual Verification:
 1. Add or edit a variable using ListAssets query type.
 2. Verify that: 
    -  the take input field is visible in the editor.
    -  default value is 1000
    -  entering invalid values (< 0, > TAKE_LIMIT) shows the correct error message in the UI and does not trigger a query
    -  valid values trigger the query and limit the results accordingly.
 3. Save dashboard and reload:
    - verify that variables without an explicit take use the default value.
 4. Edge case validation:
    - take = 0 returns all data (based on backend behavior)
    - take = TAKE_LIMIT returns the maximum allowed.
    
TakeQueryVariable

## ✅ Checklist

### DIFF
--- a/src/datasources/asset/AssetDataSource.ts
+++ b/src/datasources/asset/AssetDataSource.ts
@@ -20,9 +20,11 @@ import { ListAssetsQuery } from './types/ListAssets.types';
 import { ListAssetsDataSource } from './data-sources/list-assets/ListAssetsDataSource';
 import { AssetSummaryDataSource } from './data-sources/asset-summary/AssetSummaryDataSource';
 import { AssetModel, AssetsResponse } from 'datasources/asset-common/types';
-import { QUERY_LIMIT } from './constants/constants';
 import { transformComputedFieldsQuery } from 'core/query-builder.utils';
 import { AssetVariableQuery } from './types/AssetVariableQuery.types';
+import { defaultListAssetsVariable } from './defaults';
+import { TAKE_LIMIT } from './constants/ListAssets.constants';
+import { QUERY_LIMIT } from './constants/constants';
 
 export class AssetDataSource extends DataSourceBase<AssetQuery, AssetDataSourceOptions> {
   private assetSummaryDataSource: AssetSummaryDataSource;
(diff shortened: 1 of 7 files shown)


saradei-ni (approved):

src/datasources/asset/AssetDataSource.ts
```diff
@@ -110,4 +115,12 @@ export class AssetDataSource extends DataSourceBase<AssetQuery, AssetDataSourceO
 
     return { text: assetName, value: assetValue };
   }
+
+  public patchListAssetQueryVariable(query: AssetQuery): AssetVariableQuery {
+    return { ...defaultListAssetsVariable, ...query } as AssetVariableQuery;
+  }
+
+  private isTakeValid(take: number | undefined): boolean {
```
saradei-ni, line 123:
```suggestion
  private isTakeValid(take?: number): boolean {
```

PR #396, opened 2025-09-02
perf(asset): Update processTotalCountAssetsQuery so only one asset with one property is returned

BlankaOnodi:
# Pull Request

## 🤨 Rationale

This PR changes the variables with which we are calling queryAssets function, so only one asset with one property is returned and also the total count of assets returned.

## 👩‍💻 Implementation

In ListAssetsDataSource.ts I've updated processTotalCountAssetsQuery so:
- "response" variable contains a list of assets with only one asset with the id property selected and also the total count of assets returned
- changed "await this.queryAssets(query.filter, QUERY_LIMIT, true, query.properties)" into "await this.queryAssets(query.filter, 1, true, [AssetFilterPropertiesOption.AssetIdentifier])"
- updated "should call queryAsset with returnCount set to true when outpuType is set to TotalCount" test to call the function with take field set to 1 instead of 1000

## 🧪 Testing

Manual testing:
- Verify that:
 1. when Total Count is selected the number of assets is displayed
 2. checked the console response when Total Count is selected to display the right result

image

## ✅ Checklist

### DIFF
--- a/src/datasources/asset/data-sources/list-assets/ListAssetsDataSource.test.ts
+++ b/src/datasources/asset/data-sources/list-assets/ListAssetsDataSource.test.ts
@@ -178,7 +178,7 @@ describe('shouldRunQuery', () => {
 
         await datastore.query(query);
 
-        expect(queryAssetSpy).toHaveBeenCalledWith('', 1000, true, ["AssetIdentifier"]);
+        expect(queryAssetSpy).toHaveBeenCalledWith('', 1, true, ["AssetIdentifier"]);
     })
 
     test('should call queryAsset with returnCount set to false when outpuType is set to Properties', async () => {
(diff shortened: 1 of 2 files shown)


saradei-ni (approved):

src/datasources/asset/data-sources/list-assets/ListAssetsDataSource.test.ts
```diff
@@ -178,7 +178,7 @@ describe('shouldRunQuery', () => {
 
         await datastore.query(query);
 
-        expect(queryAssetSpy).toHaveBeenCalledWith('', 1000, true, ["AssetIdentifier"]);
```
kkerezsi, line 181: @BlankaOnodi don't you already have a constant for "AssetIdentifier" ? 

AssetFilterPropertiesOption.AssetIdentifier ? 

Normally we want to avoid hardcoding strings. Imagine if we need to change the value .. now you need to update in 1000 places . That's why we use constants, so we don't need to worry about missing renaming something.

kkerezsi (changes requested):

kkerezsi (approved):

PR #412, opened 2025-09-17
Users/ahalya/chore/alarms/cleanup

Ahalya-ni:
# Pull Request

## 🤨 Rationale

This PR cleans up the alarms datasource by updating import statements to use single quotes instead of double quotes. It also moves the `AlarmsDataSourceCore` file and its test file from the alarms root directory to the `query-type-handlers` folder, as this file is an abstract class containing methods and properties shared by various query type datasources such as `ListAlarmsDataSource`, `AlarmTrendDataSource`, and `AlarmsCountDataSource`.

## 👩‍💻 Implementation

- Updated import statements to use single quotes.
- Moved `AlarmsDataSourceCore` and its test file from the root directory to the query-type-handlers folder.

## 🧪 Testing

NA

## ✅ Checklist


Ahalya-ni:
To the reviewers, I would like to get your thoughts on this:

To ensure consistency and future-proof the codebase against accidental use of double quote styles, we could introduce the ESLint rule `"quotes": ["error", "single"]`. However, implementing this rule will require updating other affected files, specifically `querybuilder.utils.ts` and `WorkordersDataSource.ts`, along with alarms.

src/datasources/alarms/AlarmsDataSource.test.ts
```diff
@@ -17,7 +17,7 @@ describe('AlarmsDataSource', () => {
   });
 
   it('should initialize with AlarmsCount as the default query', () => {
-    expect(datastore.defaultQuery).toEqual({ queryType: QueryType.AlarmsCount });
+    expect(datastore.defaultQuery).toEqual({ queryType: QueryType.AlarmsCount, filter: '' });
```
manisha-ni: Why is this change required now? How was it passing earlier?

src/datasources/alarms/query-type-handlers/alarms-count/AlarmsCountDataSource.ts
```diff
@@ -1,10 +1,10 @@
 import { DataFrameDTO, DataQueryRequest, FieldType } from '@grafana/data';
-import { AlarmsDataSourceCore } from 'datasources/alarms/AlarmsDataSourceCore';
-import { defaultAlarmsCountQuery } from 'datasources/alarms/constants/defaultQueries';
+import { AlarmsQueryHandlersCore } from 'datasources/alarms/query-type-handlers/AlarmsQueryHandlersCore';
 import { MINIMUM_TAKE } from 'datasources/alarms/constants/QueryAlarms.constants';
 import { AlarmsCountQuery } from 'datasources/alarms/types/AlarmsCount.types';
+import { defaultAlarmsCountQuery } from 'datasources/alarms/constants/DefaultQueries.constants';
 
-export class AlarmsCountDataSource extends AlarmsDataSourceCore {
+export class AlarmsCountDataSource extends AlarmsQueryHandlersCore {
```
manisha-ni: `AlarmsCountDataSource` should be renamed to `AlarmsCountQueryHandler` right?

kartheeswaran-ni (changes requested):
I’d recommend keeping PRs focused on a single purpose, even for changes like these. This helps reviewers—and future developers—quickly understand the intent of the PR and do a much effective review.

For example:
One PR just for changing quotes
Another for updating default queries or removing an unnecessary type file
A separate one for renaming a class
Keeping changes isolated like this makes the review process smoother and faster. 
 
PR review in GitHub is not as good as AzDo. So, I would like you guys to follow these.
 
Please split and let me know. 

_Note:  I understand this particular PR isn’t very large, but I’m using it as an opportunity to encourage this process._

PR #710, opened 2026-03-23
feat(systems): Add acceptance tests for systems data source

BlankaOnodi:
# Pull Request

## 🤨 Rationale

Add acceptance tests for systems data source, which ensures user-visible functionality works as expected.
[User Story 3782470](https://dev.azure.com/ni/DevCentral/_workitems/edit/3782470): Tests for systems data source

## 👩‍💻 Implementation

Added acceptance tests for:
- properties query type, testing its integration with systems variables,
- summary query type (checking if the connected/disconnected/virtual values are displayed correctly in the table),
- systems data source configuration (also checking if the feature flag is enabled by default).

## 🧪 Testing

Ran acceptance tests locally that cover:

- systems variable integration in the systems ds
- query builder's filter affects the displayed result
- right messages are displayed when adding/deleting a data source successfully
- error message is visible when a wrong url is entered in the data source creation page
- feature flag is enabled by default

## ✅ Checklist


acceptance-tests/page-objects/dashboard/components/settings.component.ts
```diff
@@ -1,25 +1,32 @@
...
+        return this.page.getByTestId('data-testid dashboard-settings-close');
+    }
+
+    public getVariable(variableName: string): Locator {
         return this.page.getByRole('gridcell', { name: `Variable editor Table Definition field ${variableName}` });
     }
 
-    public get closeSettingsButton(): Locator {
-        return this.page.getByTestId('data-testid dashboard-settings-close');
+    public createdVariable(variableName: string) {
```
stefanc18: getVariable and createdVariable are identical

acceptance-tests/page-objects/dashboard/components/table.component.ts
```diff
@@ -11,6 +11,14 @@ export class Table {
         return this.page.getByRole('group').first();
     }
 
+    public get secondFilterRow(): Locator {
+        return this.page.getByRole('group').nth(2);
+    }
+
+    public get thirdFilterRow(): Locator {
+        return this.page.getByRole('group').nth(4);
```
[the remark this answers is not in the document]
stefanc18: It's also a bit confusing that currently thirdFilterRow selects the 4th index instead of the third one. Any reason for that?

If there really is one and we're trying to skip one, maybe we can name the method to be more suggestive or add a comment explaining why exactly we need to do that

acceptance-tests/page-objects/dashboard/components/dashboard-variable/dashboard-asset-variable.component.ts
```diff
@@ -0,0 +1,17 @@
+import { Page } from "playwright/test";
+import { DashboardVariableBaseComponent } from "./dashboard-variable-base.component";
+
+export class DashboardAssetVariableComponent extends DashboardVariableBaseComponent {
+    constructor(page: Page) {
+        super(page);
+    }
+
+    public queryReturnTypeDropdown(dropdownOptionName: string) {
```
stefanc18: This method doesn't have a return type specified.

We usually easly catch those things through lint on other UI apps, I'm curious what is the status on linting & code formatting for this project.

We don't have to change it now since enforcing stricter rules might require more changes, just wanted to bring awareness.

stefanc18 (approved):

acceptance-tests/page-objects/dashboard/components/settings.component.ts
BlankaOnodi: I just removed the "getVariable" duplicated function and changed the only place where it was used to the identical "createdVariable" function.

acceptance-tests/tests/datasources/asset/global-variable-integration/asset-ds.minion-id-return-type.spec.ts
```diff
@@ -42,17 +43,17 @@ test.describe('Asset data source with minion id return type', () => {
...
 
         test('should add filter by minionId using the asset variable', async () => {
             await dashboard.panel.assetQueryEditor.addFilter('Asset Identifier', 'equals', '$id');
 
-            await expect(dashboard.panel.table.firstFilterRow).toContainText('Asset Identifier');
-            await expect(dashboard.panel.table.firstFilterRow).toContainText('equals');
-            await expect(dashboard.panel.table.firstFilterRow).toContainText('$id');
+            await expect(dashboard.panel.table.filterRow(0)).toContainText('Asset Identifier');
+            await expect(dashboard.panel.table.filterRow(0)).toContainText('equals');
+            await expect(dashboard.panel.table.filterRow(0)).toContainText('$id');
```
saradei-ni:
```suggestion
            const filterRow = dashboard.panel.table.filterRow(0);
            await expect(filterRow ).toContainText('Asset Identifier');
            await expect(filterRow ).toContainText('equals');
            await expect(filterRow ).toContainText('$id');
```

Would something like this work as well? maybe we could use this approach on similar occurrences

saradei-ni (approved):

PR #714, opened 2026-04-08
upgrade to v12

saradei-ni:
# Pull Request

## 🤨 Rationale

Upgrade to v12

## 👩‍💻 Implementation

Upgraded packages

## 🧪 Testing

Ran acceptance tests, prepared compatibility update in another pr.

## ✅ Checklist


src/datasources/alarms/components/editors/alarm-trend/AlarmTrendQueryEditor.tsx
```diff
@@ -31,7 +31,6 @@ export function AlarmTrendQueryEditor({ query, handleQueryChange, datasource }:
       const value = (event as CustomEvent).detail.linq;
       
       if (query.filter !== value) {
-        query.filter = value;
```
richie-ni, line 34: The explicit assignment of the filter value was necessary because, without it, the query builder component treats it as another change event and causes duplicate API calls. 

Since in this newer version, reverting this would cause an ESLint error because `query.filter = value;` directly mutates the prop, which is no longer allowed .

Instead can we use a ref based variable to track the last filter value , this way we can check if we've already handled this value without mutating the prop.

Something like : 
    
    const lastFilterRef = useRef(query.filter);
     ...

    if (query.filter !== value && lastFilterRef.current !== value) {
      lastFilterRef.current = value;
      handleQueryChange({ ...query, filter: value });
    }

richie-ni: Same goes with ListAlarmsQueryEditor.tsx, AlarmsVariableQueryEditor.tsx,QueryResultsEditor.tsx & QueryStepsEditor.tsx.

richie-ni (changes requested):

src/datasources/alarms/components/editors/alarm-trend/AlarmTrendQueryEditor.tsx
saradei-ni: Thanks for pointing it out, will look into it.

PR #718, opened 2026-04-27
fix(tags): Add "tag path" field to the result table

BlankaOnodi:
# Pull Request

## 🤨 Rationale

Add a "Show tag path" toggle to the Tag datasource (Current query type) so users can include the tag path as a separate column in query results, similar to the existing Properties switch.

## 👩‍💻 Implementation

Added a showTagPath: boolean field to TagQuery. When enabled, CurrentQueryHandler appends a path field to the result data frame with each tag's raw path. The switch is hidden for History queries (same as Properties). Backwards compatible as old dashboards default to false via prepareQuery.

## 🧪 Testing

Unit tests:
-  check that the path field appears in results when enabled
- verifies the "Show path" switch renders/hides with query type changes and fires onChange with the correct value.
- Snapshots updated to reflect the new field.

tagsBug

## ✅ Checklist

### DIFF
--- a/src/datasources/tag/components/TagQueryEditor.tsx
+++ b/src/datasources/tag/components/TagQueryEditor.tsx
@@ -32,13 +32,18 @@ export function TagQueryEditor({ query, onChange, onRunQuery, datasource }: Prop
     onRunQuery();
   };
 
+  const onShowTagPathChange = () => {
+    onChange({ ...query, showTagPath: !query.showTagPath });
+    onRunQuery();
+  };
+
   return (
     <>
       <InlineField label="Query type" labelWidth={14} tooltip={tooltips.queryType}>
-        <RadioButtonGroup options={enumToOptions(TagQueryType)} value={query.type} onChange={onTypeChange}/>
+        <RadioButtonGroup options={enumToOptions(TagQueryType)} value={query.type} onChange={onTypeChange} />
       </InlineField>
       <InlineField label="Tag path" labelWidth={14} tooltip={tooltips.tagPath}>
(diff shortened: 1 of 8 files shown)


joshuaprewitt:
Looks great!

src/datasources/tag/components/TagQueryEditor.tsx
```diff
@@ -51,9 +56,14 @@ export function TagQueryEditor({ query, onChange, onRunQuery, datasource }: Prop
...
       </InlineField>
       {query.type === TagQueryType.Current && (
-        <InlineField label="Properties" labelWidth={14} tooltip={tooltips.properties}>
-          <InlineSwitch onChange={onPropertiesChange} value={query.properties}/>
-        </InlineField>
+        <>
+          <InlineField label="Properties" labelWidth={16} tooltip={tooltips.properties}>
+            <InlineSwitch onChange={onPropertiesChange} value={query.properties} />
+          </InlineField>
+          <InlineField label="Show tag path" labelWidth={16} tooltip={tooltips.showTagPath}>
```
saradei-ni: `Properties` and `Show tag path` have similar behaviors but different naming conventions. As simply going with `Tag path` would cause a conflict with the already existing field, maybe we could go for `Show properties` as well?

@kkerezsi any input on this?

kkerezsi: I agree with you. "Show Properties" and "Show tag path" works best (even though it's not consistent with the rest of the datasources). It actually is a particular usecase where we Properties for Tags means custom properties and not first class properties on the model (as it is for Assets for example)

BlankaOnodi: I changed the Properties label to Show properties.

kkerezsi (approved):

PR #719, opened 2026-04-28
fix(systems): Add "host name" field to the result table

BlankaOnodi:
# Pull Request

## 🤨 Rationale

Added "hostname" field to the Systems datasource's result table.
[Bug 3838017](https://dev.azure.com/ni/DevCentral/_workitems/edit/3838017): System Data Source does not provide the Host Name

## 👩‍💻 Implementation

Added "hostName" to the defaultProjection and the "hostname" field to the end of the result table.

## 🧪 Testing

Unit tests:
- check that "hostname" field appears in the results table

image

## ✅ Checklist


saradei-ni (approved):

src/datasources/system/datasource-query-handlers/SystemQueryHandlerBase.ts
```diff
@@ -70,7 +70,8 @@ export abstract class SystemQueryHandlerBase {
           values: properties.map(m => NetworkUtils.getIpAddressFromInterfaces(m.ip4Interfaces, m.ip6Interfaces)),
         },
         { name: 'workspace', values: properties.map(m => getWorkspaceName(workspaces, m.workspace)) },
-        { name: 'scan code', values: properties.map(m => m.scanCode) }
+        { name: 'scan code', values: properties.map(m => m.scanCode) },
+        { name: 'host name', values: properties.map(m => m.hostName) },
```
stefanc18: Let's use `hostname` instead of `host name` for the displayed string, it's consistent with the Systems UI (we even received a bug a while ago since it was 2 words there initially as well) 😃

stefanc18 (approved):

src/datasources/system/datasource-query-handlers/SystemQueryHandlerBase.ts
BlankaOnodi: I implemented the requested changes.

PR #722, opened 2026-05-06
fix(dataframe): Fix clipped dropdowns in Query configurations collapse

Ahalya-ni:
# Pull Request

## 🤨 Rationale

**Collapse `overflow` styling**: The `Collapse` component from `@grafana/ui` applies `overflow: hidden` to its body, which clips the `SlQueryBuilder` dropdowns when they open beyond the collapse body boundary.

The three query builders are stacked in order - Results → Data Table → Column. For Results and Data Table, the query builders below them extend the collapse body height enough that their dropdowns open within the visible area. The Column query builder sits at the very bottom with nothing below it, so its dropdowns are immediately clipped when they open.

## 👩‍💻 Implementation

Override `overflow: hidden` to `overflow: visible` on the `Collapse` body, scoped only to the Query configurations section so other `Collapse` sections are unaffected.

## 🧪 Testing

Manually verified and added unit test
options_dropdown_grafana

## ✅ Checklist

### DIFF
--- a/src/datasources/data-frame/components/v2/DataFrameQueryEditorV2.scss
+++ b/src/datasources/data-frame/components/v2/DataFrameQueryEditorV2.scss
@@ -1,3 +1,9 @@
+.query-configuration-section {
+    [class*='collapse__body'] {
+        overflow: visible;
+    }
+}
+
 .property-selection-section {
     [role='alert'] {
         width: 524px;
--- a/src/datasources/data-frame/components/v2/DataFrameQueryEditorV2.test.tsx
+++ b/src/datasources/data-frame/components/v2/DataFrameQueryEditorV2.test.tsx
@@ -4218,5 +4218,16 @@ describe("DataFrameQueryEditorV2", () => {
         });
     });
 
+    describe("query configuration section", () => {
+        it("should render a collapse element with a class containing 'collapse__body'", () => {
+            renderComponent();
+
+            const queryConfigSection = document.querySelector('.query-configuration-section');
+            expect(queryConfigSection).toBeInTheDocument();
+
+            const collapseBody = queryConfigSection!.querySelector('[class*="collapse__body"]');
+            expect(collapseBody).not.toBeNull();
+        });
+    });
 });
 
--- a/src/datasources/data-frame/components/v2/DataFrameQueryEditorV2.test.tsx
+++ b/src/datasources/data-frame/components/v2/DataFrameQueryEditorV2.test.tsx
@@ -4220,10 +4220,10 @@ describe("DataFrameQueryEditorV2", () => {
 
     describe("query configuration section", () => {
         it("should render a collapse element with a class containing 'collapse__body'", () => {
-            renderComponent();
+            const { renderResult } = renderComponent();
 
-            const queryConfigSection = document.querySelector('.query-configuration-section');
-            expect(queryConfigSection).toBeInTheDocument();
+            const queryConfigSection = renderResult.container.querySelector('.query-configuration-section');
+            expect(queryConfigSection).not.toBeNull();
 
             const collapseBody = queryConfigSection!.querySelector('[class*="collapse__body"]');
             expect(collapseBody).not.toBeNull();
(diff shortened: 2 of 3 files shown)


richie-ni (approved):

src/datasources/data-frame/components/v2/DataFrameQueryEditorV2.scss
```diff
@@ -1,3 +1,9 @@
+.query-configuration-section {
+    [class*='collapse__body'] {
```
kartheeswaran-ni, line 2: Fragile selector — [class*='collapse__body'] depends on Grafana's internal CSS module naming. If Grafana renames or restructures this class (e.g., in a major version bump), the fix silently breaks with no compile-time warning.

Can you try adding some unit test which will capture this class name change and fail if it's changed?

src/datasources/data-frame/components/v2/DataFrameQueryEditorV2.scss
```diff
@@ -1,3 +1,9 @@
...
+        overflow: visible;
```
kartheeswaran-ni, line 3: Comment by copilot: Collapse animation bleed-through — Grafana's [Collapse](vscode-file://vscode-app/c:/Users/karsubra/AppData/Local/Programs/Microsoft%20VS%20Code/8b640eef5a/resources/app/out/vs/code/electron-browser/workbench/workbench.html) uses overflow: hidden specifically to clip content during the open/close height transition. With overflow: visible, content inside the Query configurations collapse will visually leak outside the container during the expand/collapse animation. This produces a brief visual glitch.

Have you verified this?

src/datasources/data-frame/components/v2/DataFrameQueryEditorV2.scss
kartheeswaran-ni, line 2: Comment by Copilot: 

Unintended overflow — Any content that overflows the collapse body will now be visible, not just dropdowns. Error messages, tooltips, or wide inputs that happen to exceed the boundary will bleed out too.

kartheeswaran-ni (approved):

src/datasources/data-frame/components/v2/DataFrameQueryEditorV2.scss
Ahalya-ni: Currently, we have Take input other than querybuilder inside Collapse body. The UI looks fine.
image

image

Ahalya-ni: Added unit test

src/datasources/data-frame/components/v2/DataFrameQueryEditorV2.scss
Ahalya-ni: No glitch is identified.

Please have a look at the attached GIF.

options_dropdown_grafana_1

src/datasources/data-frame/components/v2/DataFrameQueryEditorV2.test.tsx
```diff
@@ -4218,5 +4218,16 @@ describe("DataFrameQueryEditorV2", () => {
...
 
+    describe("query configuration section", () => {
+        it("should render a collapse element with a class containing 'collapse__body'", () => {
+            renderComponent();
+
+            const queryConfigSection = document.querySelector('.query-configuration-section');
+            expect(queryConfigSection).toBeInTheDocument();
+
+            const collapseBody = queryConfigSection!.querySelector('[class*="collapse__body"]');
+            expect(collapseBody).not.toBeNull();
```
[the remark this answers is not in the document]
Ahalya-ni: Updated it to use `renderResult.container.querySelector(...)`

The `expect(queryConfigSection).not.toBeNull()` assertion before the ! already guarantees it's non-null — if it were null, the test would fail there. Using if instead would silently skip the second assertion and cause a false positive if `queryConfigSection` is null.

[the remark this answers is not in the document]
Ahalya-ni: This test intentionally doesn't verify the CSS behavior. The purpose is to act as a canary: if Grafana renames the internal `collapse__body` class label in a future version, the SCSS fix ([class*='collapse__body']) would silently stop working. This test ensures that regression is caught early.

PR #723, opened 2026-05-07
fix(query-builder): use nullish coalescing in transformBasedOnContainsOperations

richie-ni:
# Pull Request

##  🤨 Rationale

When user selects the `does not contain` operator in the query builder in any of the data source and leaves the value empty, the datasource ends up in a run time error : 

In `transformBasedOnContainsOperations`, the logical OR (`||`) operator was used to select between `negatedValue` and `positiveValue`:

```typescript
const extractedValue = negatedValue || positiveValue;
```

This caused the bug when the query contained an **empty string value** in a negated Contains operation, e.g. `!(source.Contains(""))`. Since `""` is falsy in JavaScript, `"" || undefined` evaluates to `undefined`, so the transformation function received `undefined` instead of the intended empty string.

## 👩‍💻 Implementation

Replaced `||` with `??` (nullish coalescing operator):

```typescript
const extractedValue = negatedValue ?? positiveValue;
```

The `??` operator only falls through to the right-hand side when the left-hand side is `null` or `undefined` - not when it is an empty string. This correctly preserves `""` as a valid extracted value.

## 🧪 Testing

- Verified that `!(field.Contains(""))` now passes `""` to the transformation (previously passed `undefined`).
- Verified that `field.Contains("")` still correctly passes `""`.
- Verified that non-empty values continue to work as before for both negated and positive Contains operations.

## Work Item
[Bug 3864088](https://dev.azure.com/ni/DevCentral/_workitems/edit/3864088): Dataframes Datasource | Property with `does not contain` operator errors out

##  ✅ Checklist

### DIFF
--- a/src/core/query-builder.utils.test.ts
+++ b/src/core/query-builder.utils.test.ts
@@ -79,6 +79,24 @@ describe('QueryBuilderUtils', () => {
       expect(result).toBe(query);
     });
 
+    it('should handle Contains operations with empty string value', () => {
+      const query = 'Object1.Contains("")';
+      const result = transformComputedFieldsQuery(query, computedDataFields);
+      expect(result).toBe('obj.prop1.Contains()');
+    });
+
+    it('should handle negated Contains operations with empty string value', () => {
+      const query = '!(Object1.Contains(""))';
+      const result = transformComputedFieldsQuery(query, computedDataFields);
+      expect(result).toBe('!(obj.prop1.Contains())');
+    });
+
+    it('should handle mixed Contains operations with empty and non-empty values', () => {
+      const query = 'Object1.Contains("") AND !(Object2.Contains("value"))';
+      const result = transformComputedFieldsQuery(query, computedDataFields);
+      expect(result).toBe('obj.prop1.Contains() AND !(obj.prop1.Contains(value))');
+    });
+
     it('should handle Any.Contains operations correctly with computed fields', () => {
       const query = 'Object1.Any(it.Contains("value1"))';
       const result = transformComputedFieldsQuery(query, computedDataFields);
--- a/src/core/query-builder.utils.test.ts
+++ b/src/core/query-builder.utils.test.ts
@@ -109,6 +109,24 @@ describe('QueryBuilderUtils', () => {
       expect(result).toBe(query);
     });
 
+    it('should handle Any.Contains operations with empty string value', () => {
+      const query = 'Object1.Any(it.Contains(""))';
+      const result = transformComputedFieldsQuery(query, computedDataFields);
+      expect(result).toBe('Object1.Any(obj.prop1.Contains())');
+    });
+
+    it('should handle negated Any.Contains operations with empty string value', () => {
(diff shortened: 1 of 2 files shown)


src/core/query-builder.utils.test.ts
```diff
@@ -79,6 +79,24 @@ describe('QueryBuilderUtils', () => {
...
+      const query = 'Object1.Contains("")';
+      const result = transformComputedFieldsQuery(query, computedDataFields);
+      expect(result).toBe('obj.prop1.Contains()');
+    });
+
+    it('should handle negated Contains operations with empty string value', () => {
+      const query = '!(Object1.Contains(""))';
+      const result = transformComputedFieldsQuery(query, computedDataFields);
+      expect(result).toBe('!(obj.prop1.Contains())');
+    });
```
kartheeswaran-ni, line 92: Should this be considered as an invalid state?

For example, in SLE Products app, empty string is considered as invalid value for the "Contains" operator.

image

kartheeswaran-ni (changes requested):

src/core/query-builder.utils.test.ts
richie-ni: The current implementation of the SlQueryBuilder does not use the validation as in our other SLE apps and hence doesn't have the invalid state. It would require changes to the SlQueryBuilder . Do you suggest we change the SlQueryBuilder to include the invalid state.

src/core/query-builder.utils.ts
```diff
@@ -66,7 +66,7 @@ function transformBasedOnContainsOperations(query: string, field: string, transf
 
   return query.replace(containsRegex, (_match, _negatedMatch, negatedValue, _positiveMatch, positiveValue) => {
     const isNegated = negatedValue !== undefined;
-    const extractedValue = negatedValue || positiveValue;
+    const extractedValue = negatedValue ?? positiveValue;
```
kartheeswaran-ni, line 69: Don't we need to fix this in `transformListContainsOperations`?

kartheeswaran-ni (approved):

src/core/query-builder.utils.test.ts
richie-ni: If yes, we can have it as a separate PR later and need not be linked to this Bug. Created a [Technical Debt 3869323](https://dev.azure.com/ni/DevCentral/_workitems/edit/3869323): Add validation for Grafana datasource query builders

src/core/query-builder.utils.ts
richie-ni: Thats a good catch, we have the same bug in the list as well.. I have added the fix to the List operators as well..

src/core/query-builder.utils.test.ts
richie-ni: Resolving this right now, please feel free to reopen in case of any other concerns.

richie-ni:
@ni/grafana-codeowners-team ,  a gentle reminder to review this PR.

saradei-ni (approved):

PR #724, opened 2026-05-07
fix(data-frame): strip CSV apostrophe escape prefix from string column values

richie-ni:
# Pull Request

## 🤨 Rationale

The undecimated data table query feature fetches raw data via the Data Frame Service's CSV export route (`POST /tables/{tableId}/export-data` with `responseFormat: 'CSV'`). The DFS backend has a CSV injection mitigation that prefixes string column values with an apostrophe (`'`) when they start with certain characters (`=`, `@`, `+`, `-`, tab, or carriage return) to prevent attacks when the CSV is opened in Excel.

The Grafana plugin was displaying these escaped values as-is (e.g., `'-5` instead of `-5`), which breaks data transforms in dashboards that expect numeric-like values and shows incorrect data to users.

Related work item: [Bug 3832031](https://dev.azure.com/ni/DevCentral/_workitems/edit/3832031)

## 👩‍💻 Implementation

Added a `stripCsvEscapePrefix` method to `DataFrameDataSourceV2` that strips the leading apostrophe **only when it is followed by one of the known DFS CSV injection trigger characters** (`=`, `@`, `+`, `-`, `\t`, `\r`). This ensures:

- Escaped values are correctly unescaped (e.g., `'-5` → `-5`)
- Strings that legitimately start with an apostrophe followed by a non-trigger character are preserved unchanged (e.g., `'hello` stays `'hello`)

The `transformValue` method now calls `stripCsvEscapePrefix` for STRING columns instead of returning the raw value directly.

**Files changed:**
- `src/datasources/data-frame/datasources/v2/DataFrameDataSourceV2.ts` — added `stripCsvEscapePrefix` and updated `transformValue`
- `src/datasources/data-frame/datasources/v2/DataFrameDataSourceV2.test.ts` — added 11 unit tests

## 🧪 Testing

- Added 11 unit tests covering:
  - All 6 CSV injection trigger characters are correctly stripped
  - Apostrophe followed by a regular character is preserved
  - Single apostrophe string is preserved
  - Null and empty string inputs are handled safely
  - Strings not starting with an apostrophe pass through unchanged
- All existing tests continue to pass

## ✅ Checklist

### DIFF
--- a/src/datasources/data-frame/datasources/v2/DataFrameDataSourceV2.ts
+++ b/src/datasources/data-frame/datasources/v2/DataFrameDataSourceV2.ts
@@ -1486,11 +1486,17 @@ export class DataFrameDataSourceV2 extends DataFrameDataSourceBase {
     }
 
     private stripCsvEscapePrefix(value: string | null): string | null {
-        if (!value || value.length < 2 || value[0] !== "'") {
+        const csvInjectionTriggers = ['=', '@', '+', '-', '\t', '\r'];
+        if (
+            !value
+            || value.length < 2
+            || value[0] !== "'"
+            || !csvInjectionTriggers.includes(value[1])
+        ) {
             return value;
         }
-        const csvInjectionTriggers = ['=', '@', '+', '-', '\t', '\r'];
-        return csvInjectionTriggers.includes(value[1]) ? value.substring(1) : value;
+
+        return value.substring(1);
     }
 
     private getFieldTypeForDataType(dataType: ColumnDataType): FieldType {
--- a/src/datasources/data-frame/datasources/v2/DataFrameDataSourceV2.ts
+++ b/src/datasources/data-frame/datasources/v2/DataFrameDataSourceV2.ts
@@ -1481,8 +1481,16 @@ export class DataFrameDataSourceV2 extends DataFrameDataSourceBase {
             case 'TIMESTAMP':
                 return dateTime(value).valueOf();
             default:
-                return value;
+                return this.stripCsvEscapePrefix(value);
         }
+    }
+
+    private stripCsvEscapePrefix(value: string | null): string | null {
+        if (!value || value.length < 2 || value[0] !== "'") {
+            return value;
+        }
+        const csvInjectionTriggers = ['=', '@', '+', '-', '\t', '\r'];
+        return csvInjectionTriggers.includes(value[1]) ? value.substring(1) : value;
     }
 
     private getFieldTypeForDataType(dataType: ColumnDataType): FieldType {
--- a/src/datasources/data-frame/datasources/v2/DataFrameDataSourceV2.test.ts
+++ b/src/datasources/data-frame/datasources/v2/DataFrameDataSourceV2.test.ts
@@ -11835,4 +11835,61 @@ describe('DataFrameDataSourceV2', () => {
             });
         });
     });
+
+    describe('stripCsvEscapePrefix', () => {
+        it('should strip apostrophe prefix when followed by equals sign', () => {
+            const result = (ds as any).stripCsvEscapePrefix("'=SUM(A1)");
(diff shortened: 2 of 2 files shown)


src/datasources/data-frame/datasources/v2/DataFrameDataSourceV2.ts
```diff
@@ -1481,10 +1481,18 @@ export class DataFrameDataSourceV2 extends DataFrameDataSourceBase {
...
         }
     }
 
+    private stripCsvEscapePrefix(value: string | null): string | null {
+        if (!value || value.length < 2 || value[0] !== "'") {
+            return value;
+        }
+        const csvInjectionTriggers = ['=', '@', '+', '-', '\t', '\r'];
+        return csvInjectionTriggers.includes(value[1]) ? value.substring(1) : value;
+    }
```
kartheeswaran-ni, line 1500: Question about stripCsvEscapePrefix behavior for edge cases:

What is the expected behavior when a user has intentionally stored a value that starts with a single quote followed by a trigger character — e.g., the literal string '=20?

Currently:

User stores =20 → backend escapes to '=20 → plugin strips ' → displays =20 ✓

User stores '=20 → backend sends '=20 as-is (since ' is not in the trigger list) → plugin sees ' + = → incorrectly strips the user's intentional quote → displays =20 instead of '=20 ✗

This is ambiguous because the plugin cannot distinguish between:

A backend-escaped value (original was =20, backend added ' prefix)

A user-stored value that genuinely starts with ' followed by a trigger char

Could you align with the backend team on the expected contract here?

kartheeswaran-ni (changes requested):

src/datasources/data-frame/datasources/v2/DataFrameDataSourceV2.ts
richie-ni: @spanglerco, 
What do you think about this case? Would you suggest stripping the value off?

Note: As discussed in the bug, the better long-term solution would be to handle this on the server side, such as by adding "format": "CSV_RAW_STRINGS" (or something similar) to the export request body so the server skips escaping entirely. However, that would require a service-level change. This would just be a short-term workaround.

spanglerco: String columns on a dashboard already feels like a corner case to me, so the additional edge case of a pre-escaped formula in the value doesn't sound like something we need to worry about ahead of time. There isn't any plan to add a `CSV_RAW_STRINGS` format unless a PO steps in and decides that's required to fix this bug. A better option is a binary export feature (specifically for this use case, Apache Arrow), which is on the backlog but not scheduled.

spanglerco (approved):

src/datasources/data-frame/datasources/v2/DataFrameDataSourceV2.ts
```diff
@@ -1481,10 +1481,18 @@ export class DataFrameDataSourceV2 extends DataFrameDataSourceBase {
...
+                return this.stripCsvEscapePrefix(value);
         }
     }
 
+    private stripCsvEscapePrefix(value: string | null): string | null {
+        if (!value || value.length < 2 || value[0] !== "'") {
+            return value;
+        }
+        const csvInjectionTriggers = ['=', '@', '+', '-', '\t', '\r'];
+        return csvInjectionTriggers.includes(value[1]) ? value.substring(1) : value;
```
kartheeswaran-ni:
```suggestion
        const csvInjectionTriggers = ['=', '@', '+', '-', '\t', '\r'];
        if (
            !value
            || value.length < 2
            || value[0] !== "'"
            || !csvInjectionTriggers.includes(value[1])
        ) {
            return value;
        }

        return value.substring(1);
```

kartheeswaran-ni (approved):

PR #729, opened 2026-05-29
fix(dataframe): Remove highResolutionZoom feature flag

Ahalya-ni:
# Pull Request

## 🤨 Rationale
[Task 3897394](https://dev.azure.com/ni/DevCentral/_workitems/edit/3897394): Remove ff from plugins repo

The `highResolutionZoom` feature flag was introduced to gate the "filter by x-axis range on zoom or pan" behavior. The feature is now stable and ready to be permanently enabled, so the flag is being removed to simplify the codebase.

## 👩‍💻 Implementation

Removed all references to the `highResolutionZoom` feature flag across the data-frame datasource:

- **`types.ts`** - Removed `DataFrameFeatureToggles` interface, `DataFrameFeatureTogglesDefaults` constant, and `featureToggles` field from `DataFrameDataSourceOptions`.
- **`DataFrameDataSourceV2.ts`** - Removed `isHighResolutionZoomFeatureEnabled` field and its guard conditions. The x-axis range filtering and `updateSyncXAxisRangeTargetsQueryParam` call are now always active.
- **`DataFrameConfigEditor.tsx`** — Removed the feature toggles UI section (the "Filter by x-axis range on zoom or pan" switch).
- **`DataFrameQueryEditorV2.tsx`** - Removed the `isHighResolutionZoomFeatureEnabled` memo; the `filterXRangeOnZoomPan` label and tooltip are now used unconditionally.

## 🧪 Testing

- Removed unit tests that validated the disabled-flag behavior.

## ✅ Checklist


src/datasources/data-frame/datasources/v2/DataFrameDataSourceV2.test.ts
```diff
@@ -3686,17 +3541,12 @@ describe('DataFrameDataSourceV2', () => {
...
-                }
                 const undecimatedInstanceSettings = {
                     id: 1,
                     name: 'test',
                     type: 'test',
                     url: 'http://localhost',
-                    jsonData: {
-                        featureToggles
-                    }
+                    jsonData: {}
```
richie-ni, line 3549: Is an empty object still required

src/datasources/data-frame/components/DataFrameConfigEditor.tsx
```diff
@@ -1,51 +1,17 @@
...
 export const DataFrameConfigEditor: React.FC<Props> = ({ options, onOptionsChange }) => {
-    const handleFeatureChange = (featureKey: string) => (event: ChangeEvent<HTMLInputElement>) => {
-        const jsonData = {
-            ...options.jsonData,
-            featureToggles: { ...options.jsonData.featureToggles, [featureKey]: event.target.checked },
-        };
-        onOptionsChange({ ...options, jsonData });
-    };
-
     return (
```
richie-ni: since config editor only contains the HTTP setting, we could set the http config editor directly in the modules file

richie-ni (changes requested):
Can we remove the Config Editors and add the HTTP Config editor directly in the Modules file

src/datasources/data-frame/components/DataFrameConfigEditor.tsx
Ahalya-ni: Retaining the DataFrameConfigEditor wrapper for consistency with the query and variable editor wrappers - if we need to add additional config settings or handle migration in the future, it would be useful to have it in place.

Reference PR discussion - [fix(dataframe): Remove queryByDataTableProperties and queryByResultAndColumnProperties feature flags…](https://github.com/ni/systemlink-grafana-plugins/pull/658#discussion_r2739572334)

richie-ni: I do not think the config editor really helps in any kind of migration and we have already verified that the config editor's feature flag editor also doesn't help users to toggle the feature flag , hence i do not see a use of the config editor at all.
>  if we need to add additional config settings or handle migration in the future,

What config settings are we envisioning here?

Ahalya-ni: I meant adding feature flag in the config editor.

Removed config editor file and added `HttpConfigEditor` directly to the `module.ts`

src/datasources/data-frame/datasources/v2/DataFrameDataSourceV2.test.ts
Ahalya-ni: `jsonData` is a required field in `DataSourceInstanceSettings`. While the `as any` cast lets us bypass the type check, keeping `jsonData: {}` as an empty object serves as a placeholder if any future options (like feature flags) need to be added.

richie-ni (approved):

src/datasources/data-frame/components/v2/DataFrameQueryEditorV2.tsx
```diff
@@ -751,9 +746,9 @@ export const DataFrameQueryEditorV2: React.FC<Props> = (
                             />
                         </InlineField>
                         <InlineField
-                            label={isHighResolutionZoomFeatureEnabled ? labels.filterXRangeOnZoomPan : labels.useTimeRange}
+                            label={labels.filterXRangeOnZoomPan}
```
kartheeswaran-ni, line 749: Can we remove the respective constants as well if they are no longer needed?

src/datasources/data-frame/constants/v2/DataFrameQueryEditorV2.constants.ts

kartheeswaran-ni (approved):

src/datasources/data-frame/components/v2/DataFrameQueryEditorV2.tsx
Ahalya-ni: Removed `useTimeRange` label and tooltip as it is no longer needed after removing the `isHighResolutionZoomFeatureEnabled` feature flag

